PR #10, opened 2016-07-07
Add Scalar template argument to UnalignedEquivalent struct.

aelkhour:
* It allows creating unaligned equivalent of Eigen matrix using a
  different scalar type. This can be useful when matrix types use
  single precision floats, whereas Numpy matrix types use double
  precision floats by default.
* Default template argument value is Matrix D scalar type to preserve
  backward-compatibility.

### DIFF
--- a/src/fwd.hpp
+++ b/src/fwd.hpp
@@ -22,11 +22,11 @@
 
 namespace eigenpy
 {
-  template<typename D>
+  template<typename D, typename Scalar = typename D::Scalar>
   struct UnalignedEquivalent
   {
     typedef Eigen::MatrixBase<D> MatType;
-    typedef Eigen::Matrix<typename D::Scalar,
+    typedef Eigen::Matrix<Scalar,
              D::RowsAtCompileTime,
              D::ColsAtCompileTime,
 #ifndef EIGENPY_ALIGNED


jcarpent:
Hello Antonio. Nice suggestions.

Currently, your commit will affect the bindings of Pinocchio for the Spatial quantities because they do not have Scalar defined in their public typedef.
Can you also propagate it to Pinocchio?

jcarpent:
Just to inform you that I'm currently rewriting the eigenpy to fully expose Eigen quantities and not converting Eigen objects into bumpy objects. The idea is to fully exploit the Eigen capacities. It will be then possible to convert Eigen python object into bumpy objects if needed.

aelkhour:
Sorry about that @jcarpent, I forgot this was used on objects other than Eigen matrices.

If I understood correctly, you suggest then that I define Scalar for spatial quantities in pinocchio?

aelkhour:
Done in https://github.com/stack-of-tasks/pinocchio/pull/224

@jcarpent After you merge, would it be possible to make a new release of eigenpy?

jcarpent:
I need to merge some stuff about the packaging of eigenpy before the next fix release.
I will try to do my best, so let's say asap.

aelkhour:
OK thanks.

PR #14, opened 2017-01-05
Makes eigenpy compliant with NUMPY 1.8 (on Ubuntu 14.04 LTS)

olivier-stasse:
This should remove the deprecated messages spanning all over the packages relying on pinocchio and thus on eigenpy.

### DIFF
--- a/src/fwd.hpp
+++ b/src/fwd.hpp
@@ -19,7 +19,10 @@
 
 #include <boost/python.hpp>
 #include <Eigen/Core>
+
+#ifdef NPY_1_8_API_VERSION
 #define NPY_NO_DEPRECATED_API NPY_1_7_API_VERSION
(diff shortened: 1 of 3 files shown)


jcarpent:
Thanks @olivier-stasse for your contribution. Before merging, can you make your PR on the level branch?

src/fwd.hpp
```diff
@@ -19,6 +19,7 @@
 
 #include <boost/python.hpp>
 #include <Eigen/Core>
+#define NPY_NO_DEPRECATED_API NPY_1_7_API_VERSION
```
olivier-stasse: After reading the code of numpy v1.11 it seems that this not even needed.
Only NPY_1_7_API_VERSION is required.

PR #28, opened 2018-04-16
Fix bug when converting matrix of dim 1x1

jcarpent:


jcarpent:
This bug is linked to stack-of-tasks/pinocchio#453.

PR #39, opened 2018-09-13
Add note for compiling eigenpy with python3 on macOS

jviereck:
Getting eigenpy (and pinocchio) to compile on macOS with python3 took me some effort. The required trick turned out to "only" add `-DPYTHON_EXECUTABLE=\`which python3\`` to the cmake argument.

Do you think this is worth adding to the README.md/installation instructions? I am happy to provide something similar for pinocchio as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,6 +23,23 @@ directory:
 Please note that CMake produces a `CMakeCache.txt` file which should
 be deleted to reconfigure a package from scratch.
 
+#### Compiling for python3 on macOS
+
+Make sure you have boost for python3 installed. If you use homebrew, you can install it via `brew install boost-python3`.
+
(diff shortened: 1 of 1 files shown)


jcarpent:
@jviereck I just add the formula `eigenpy-python3`to our homebrew repository (https://github.com/Gepetto/homebrew-gepetto).
It is now easy to install eigenpy with python3 compatibility just by running:
```
brew tap gepetto/homebrew-gepetto
brew install eigenpy-python3
```
Best,

Justin

jviereck:
@jcarpent, fantastic! Is there a similar `apt-get install` for linux users? Then I guess adding an `installation` section to eigenpy Readme makes sense (happy to add it myself).

Issue #52, opened 2019-02-18
git submodules contain SSH-path


markusgft:
The URL url = git://github.com/jrl-umi3218/jrl-cmakemodules.git
limits accessibility for many users. 
Please apply same modifications as in https://github.com/stack-of-tasks/pinocchio/issues/667

jcarpent:
Yes, sure. I will do that now. Thanks for the quick report.

jcarpent:
Solved by #53.

PR #53, opened 2019-02-18
submodules: use https address

jcarpent:


jcarpent:
Related to #52.

.gitmodules
```diff
@@ -1,6 +1,6 @@
 [submodule "cmake"]
 	path = cmake
-	url = git://github.com/jrl-umi3218/jrl-cmakemodules.git
+	url = https://github.com/jrl-umi3218/jrl-cmakemodules.git
```
nim65s: why not `url = ../../jrl-umi3218/jrl-cmakemodules`, as this was done in pinocchio ?

jcarpent: Because it will only work if you have a similar tree of repositories as on github or gitlab. 
I think this approach is not as generic as expected.

jcarpent: I will recover the previous behavior on Pinocchio.
@nim65s Do you agree with this?

nim65s: I mirrored https://github.com/jrl-umi3218/jrl-cmakemodules to https://gepgitlab.laas.fr/jrl-umi3218/jrl-cmakemodules for this exact reason :)
But it is a bit outdated now, I will push again.

nim65s: Done, does this fix your case ?

jcarpent: What I meant is that if you are working with your clone of these repositories which are not located on the same root tree, then it won't work.
For instance, if you have Pinocchio on your own local gitlab server, but you don't want to also have the cmake module, then the current relative url won't work and will break the cloning.

nim65s: That's true. But I don't get why you would not want to just mirror jrl-umi3218/jrl-cmakemodules as I did on gepgitlab. If you are able to mirror stack-of-tasks/eigenpy, it is exactly the same process.

If you really want to impose one protocol, then HTTPS is probably a better choice than SSH, but it will upset users for which SSH would work and not HTTPS.

jcarpent: So, let's choose `https` to cover the maximum number of users.
There is also an interesting point here: https://docs.gitlab.com/ee/ci/git_submodules.html

Issue #35, opened 2018-05-28
Python 3 segfaults on exit interpreter


nim65s:
Hi,

As shown by #34, we have random segfaults on python3:
https://gepgitlab.laas.fr/gsaurel/eigenpy/-/jobs/2812

Here is a minimal working exemple triggering this segfault:

```py
import matrix
import numpy as np

m = np.reshape(range(64), (8, 8))
a = matrix.reflex(m, False)
print('end of script')
```

And here is what gdb finds:

```
(gdb) run python/test_mwe_segfault.py
Starting program: /usr/bin/python3 python/test_mwe_segfault.py
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
[New Thread 0x7ffff2f16700 (LWP 25047)]
[New Thread 0x7ffff0715700 (LWP 25048)]
[New Thread 0x7fffeff14700 (LWP 25049)]
[New Thread 0x7fffeb713700 (LWP 25050)]
[New Thread 0x7fffe8f12700 (LWP 25051)]
[New Thread 0x7fffe6711700 (LWP 25052)]
[New Thread 0x7fffe5f10700 (LWP 25053)]
end of script

Thread 1 "python3" received signal SIGSEGV, Segmentation fault.
0x00000000005a0ee8 in dict_dealloc.lto_priv.164 (mp=0x7ffff6955e48) at ../Objects/dictobject.c:1595
1595	../Objects/dictobject.c: Aucun fichier ou dossier de ce type.
(gdb) bt
#0  0x00000000005a0ee8 in dict_dealloc.lto_priv.164 (mp=0x7ffff6955e48) at ../Objects/dictobject.c:1595
#1  0x0000000000597903 in module_dealloc.lto_priv () at ../Objects/moduleobject.c:638
#2  0x00007ffff64e3d57 in eigenpy::PyMatrixType::~PyMatrixType() ()
   from /local/users/gsaurel/stack-of-tasks/eigenpy/libeigenpy.so
#3  0x00007ffff7829ff8 in __run_exit_handlers (status=0, listp=0x7ffff7bb45f8 <__exit_funcs>, 
    run_list_atexit=run_list_atexit@entry=true) at exit.c:82
#4  0x00007ffff782a045 in __GI_exit (status=<optimized out>) at exit.c:104
[...]

PR #56, opened 2019-03-09
update CI & Tests

nim65s:
Hi,

This PR updates gitlab's CI to use a generated configuration. This allows us to ease the CI configuration without having to make changes in this repository again.

Currently, this gives this result:
https://gepgitlab.laas.fr/gsaurel/eigenpy/pipelines/3452/builds

Builds on 14.04 / 16.04 on python 3 are allowed to fail because of #35. 
Some other builds are currently allowed to fail, but as they succeed, the buildfarm will automatically understand that those builds are passing, and therefore they should not be allowed to fail again.

While here, I'm fixing a PYTHONPATH for one test.


unittest/CMakeLists.txt
```diff
@@ -35,4 +35,4 @@ ENDIF()
 
 ADD_PYTHON_UNIT_TEST("py-matrix" "unittest/python/test_matrix.py" "unittest")
 ADD_PYTHON_UNIT_TEST("py-geometry" "unittest/python/test_geometry.py" "unittest")
-ADD_PYTHON_UNIT_TEST("py-switch" "unittest/python/test_switch.py" "unittest")
+ADD_PYTHON_UNIT_TEST("py-switch" "unittest/python/test_switch.py" "python/eigenpy")
```
jcarpent: Why you have to do that here?

nim65s: Because I was lazy :)

Issue #55, opened 2019-03-08
Remove dependency of libboost-python from libeigenpy.sp


jviereck:
This issue is about supporting python2 and python3 installation of eigenpy.

When compiling and installing `eigenpy`, there are two libraries installed (assuming `prefix=/opt/openrobots` here):

- /opt/openrobots/libs/libeigenpy.so
- /opt/openrobots/lib/python3.5/site-packages/eigenpy.so

Note that the first one has a dependency on the libboost_python lib, in my example

``` bash
$ ldd libeigenpy.so 
	linux-vdso.so.1 =>  (0x00007fffecf99000)
	libboost_python-py35.so.1.58.0 => /usr/lib/x86_64-linux-gnu/libboost_python-py35.so.1.58.0 (0x00007f3e4e3c4000)
	libstdc++.so.6 => /usr/lib/x86_64-linux-gnu/libstdc++.so.6 (0x00007f3e4e042000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f3e4dd39000)
	libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007f3e4db23000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f3e4d759000)
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f3e4d53c000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f3e4e8ca000)
```

(note the second line).

This dependency on `libboost_python` from `libeigenpy.so` creates troubles when trying to install eigenpy for python2 and python3 at the same time in the same folder.

As a workaround, I started to use a `/opt/openrobots` and `/opt/openrobots_py35` folder to separate the two installations of `eigenpy`. However I am wondering if there is a more proper way to resolve this issue and wanted to report this in case someone else comes along the same issue.

jcarpent:
In fact, we should have two separate libraries. One for python2 and the one for python3 in /opt/openrobots/libs. Which means that we just need to add an extended name for each lib according to the version of Python.

jcarpent:
Solved by #57

PR #57, opened 2019-03-09
Set the main lib name according to PEP 3149

nim65s:
Hi,

This PR implements https://github.com/jrl-umi3218/jrl-cmakemodules/pull/177 for the main library, sot that we will generate:
- `lib/libeigenpy.so` (linked to boost-python 2) and `lib/python2.7/site-packages/eigenpy.so` for python 2
- `lib/libeigenpy.cpython-37m-x86_64-linux-gnu.so` (linked to boost-python 3) and `lib/python3.7/site-packages/eigenpy.so` for python 3 (this obviously depends on the actually used python version / architecture / etc).

This should fix the issue raised in #55, ie. allows us to install the python 2 and python 3 libraries side by side in the same prefix.

But to achieve this goal, we should also handle how the other files (includes and .pc) are installed. I will provide a solution later.


nim65s:
CI including #56 & this PR is provided here: https://gepgitlab.laas.fr/gsaurel/eigenpy/pipelines/3453

jcarpent:
@nim65s Thanks for this contribution

PR #58, opened 2019-04-11
Fix SEGV with Python 3.

jmirabel:
Related to #35 

I don't know exactly why this works (and whether this works for everyone).

### DIFF
--- a/include/eigenpy/details.hpp
+++ b/include/eigenpy/details.hpp
@@ -8,6 +8,7 @@
 
 #include "eigenpy/fwd.hpp"
 
+#include <patchlevel.h> // For PY_MAJOR_VERSION
 #include <numpy/arrayobject.h>
 #include <iostream>
 
@@ -93,9 +94,11 @@ namespace eigenpy
     NumpyType()
     {
       pyModule = bp::import("numpy");
+#if PY_MAJOR_VERSION >= 3
       // TODO I don't know why this Py_INCREF is necessary.
       // Without it, the destructor of NumpyType SEGV sometimes.
       Py_INCREF(pyModule.ptr());
+#endif
       
       NumpyMatrixObject = pyModule.attr("matrix");
       NumpyMatrixType = reinterpret_cast<PyTypeObject*>(NumpyMatrixObject.ptr());
--- a/include/eigenpy/details.hpp
+++ b/include/eigenpy/details.hpp
@@ -93,11 +93,12 @@ namespace eigenpy
     NumpyType()
     {
       pyModule = bp::import("numpy");
+      // TODO I don't know why this Py_INCREF is necessary.
+      // Without it, the destructor of NumpyType SEGV sometimes.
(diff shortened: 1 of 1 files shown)


nim65s:
I can confirm that this fix SEGV on Arch (boost 1.69, python3.7 & numpy 1.16.2)

nim65s:
It also works on 18.04. On both Arch & 18.04, the `switch` test is not passing, but this is another issue for which I can provide a fix.

nim65s:
https://gepgitlab.laas.fr/gsaurel/eigenpy/pipelines/4046 LGTM !

It would be really nice if we could make a release with this, so that I can ship it with pinocchio v2.1.2 & HPP 4.5 on my next robotpkg PR

jmirabel:
I ended up with this fix because the SEGV arises when deleting the numpy module object. I guess it is deleted twice.

Increasing the reference count may have the side effect of Numpy module not being deleted. I didn't check that. It shouldn't be hard with a break point in gdb.

include/eigenpy/details.hpp
```diff
@@ -93,11 +93,12 @@ namespace eigenpy
     NumpyType()
     {
       pyModule = bp::import("numpy");
+      // TODO I don't know why this Py_INCREF is necessary.
+      // Without it, the destructor of NumpyType SEGV sometimes.
+      Py_INCREF(pyModule.ptr());
       
       NumpyMatrixObject = pyModule.attr("matrix");
       NumpyMatrixType = reinterpret_cast<PyTypeObject*>(NumpyMatrixObject.ptr());
-      NumpyAsMatrixObject = pyModule.attr("asmatrix");
```
jcarpent: Why have you deleted that? Is it deprecated on newer versions of NumPy?

jcarpent:
@jmirabel I indeed that maybe it may happen that some data are kept in memory with this increase of reference. Maybe, we should check the Python version and not apply it to Python 2 & 3.

include/eigenpy/details.hpp
jmirabel: It isn't used so I thought it was here as a result of development code not being cleaned up.

jmirabel:
I am not gonna do a specific case for Python 2. I would rather say: for Python 2, use eigenpy <= 1.5.

We can open an issue in case someone is able to spend time on it.

include/eigenpy/details.hpp
jcarpent: I don’t agree as part of the current users are still using Python 2. We cannot deprecate now this feature even if Python will be over soon.

jcarpent:
I don’t agree as part of the current users are still using Python 2. We cannot deprecate now this feature even if Python will be over soon.

include/eigenpy/details.hpp
jcarpent: I would prefer to keep as they are not related to the current bug.

PR #59, opened 2019-04-16
Restoring previous removed code

jcarpent:


jmirabel:
For your info, 657208e makes the SEGV on exit reappearing. Running script from #35 several yields a SEGV from time to time (3 out of 10 in my case).

jcarpent:
@jmirabel The behavior you mentioned is really weird, don't you think so?
Anyway, I will just comment those lines, hoping that the behavior you've mentioned will disappear.

jmirabel:
Yes, it is. Something does bad things with the references on some part of NumPy. I have no idea how to find who... Maybe valgrind can help us.

jcarpent:
By the way, I did not get this behavior on my computer. Are sure that you're working with the latest devel version?

jmirabel:
Yes, I am sure. With Python 3.5.2 and NumPy 1.11.0.
I have to run it several times to get a SEGV. It is not consistent over
runs.
```python
import numpy, matrix
import numpy as np

m = np.reshape(range(64), (8, 8))
a = matrix.reflex(m, False)
print('end of script')
```

jcarpent:
I've tried several times and I don't get this issue with numpy 1.16 and Python 3.7.3

jmirabel:
Me too as mentioned here: https://github.com/stack-of-tasks/eigenpy/issues/35#issuecomment-482041591

jcarpent:
Yes, but without your fix, it was not working. I had a segfault with Python 3.7.3 and NumPy 1.16.
But, I don't see your point concerning the restore commit of "asmatrix".

jmirabel:
My point is that I have a SEGV with 657208e and I don't have a SEGV when I revert it.

Issue #63, opened 2019-06-23
eigenpy breaks array reshaping


rhaschke:
When using eigenpy, flattening of arrays fails:
```python
import numpy
v = numpy.array([1,2,3,4]).reshape(-1,1)  # column vector
print v.flatten().shape  # should result in (4,), but yields (4,1) when using eigenpy
```

jcarpent:
I will have a look at it. Thanks for raising this bug.

jcarpent:
I've tried on my computer, and the bug you mentioned is not occurring. Can you provide me your NumPy version and the eigenpy version you're currently using?

rhaschke:
numpy: 1.13.3
eigenpy: 1.5.1

But, I guess you were able to reproduce the problem?
I can only validate your fix tomorrow... Thanks for the quick response.

rhaschke:
With your fix my particular problem still exists. I have created a minimal example to reproduce the issue:
```python
from __future__ import print_function

import eigenpy
import numpy as np

quat = eigenpy.Quaternion(1,2,3,4)
a = quat.coeffs().T
print(a, a.flatten())

b = np.array([[1.,2.,3.,4.]])
print(b, b.flatten())
```

which yields:
```
[[ 2.  3.  4.  1.]] [[ 2.  3.  4.  1.]]
[[ 1.  2.  3.  4.]] [ 1.  2.  3.  4.]
```
You can see that the `flatten()` on the Eigen matrix doesn't yield a one-dimensional array, but it stays to be a 1xN matrix.

rhaschke:
Using `np.asarray(a).flatten()` as a workaround, my issue is solved now.
When I was prototyping my code two days ago, I didn't notice that eigenpy objects cannot handle `flatten()` in the usual way and that drove me nearly crazy because I got mismatching dimensions all the time...

jcarpent:
You should in fact use those lines:
```
from __future__ import print_function

import eigenpy
eigenpy.switchToNumpyArray() # Allows to convert Eigen::Matrix into np.array objects
import numpy as np

quat = eigenpy.Quaternion(1,2,3,4)
a = quat.coeffs().T
print(a, a.flatten())

b = np.array([[1.,2.,3.,4.]])
print(b, b.flatten())
```

The content summarizes the way it works. Hope it will help you.

rhaschke:
[...]

PR #64, opened 2019-06-24
core: fix dimensions of vecrtor for Array conversion

jcarpent:
Related to #63.
cc @rhaschke seems to solve your issue.

### DIFF
--- a/include/eigenpy/details.hpp
+++ b/include/eigenpy/details.hpp
@@ -44,6 +44,12 @@ namespace eigenpy
 
   namespace bp = boost::python;
 
+  enum NP_TYPE
+  {
+    MATRIX_TYPE,
+    ARRAY_TYPE
+  };
+  
   struct NumpyType
   {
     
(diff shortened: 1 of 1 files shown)


include/eigenpy/details.hpp
```diff
@@ -117,8 +130,12 @@ namespace eigenpy
     bp::object NumpyMatrixObject; PyTypeObject * NumpyMatrixType;
     //bp::object NumpyAsMatrixObject; PyTypeObject * NumpyAsMatrixType;
     bp::object NumpyArrayObject; PyTypeObject * NumpyArrayType;
+    
+    static NP_TYPE np_type;
```
rhaschke: I'm not deep into the code. But storing the type (array/matrix) statically doesn't seem to make sense for me on first sight.
Can we still have multiple `ndarray` instances with different type then?

jcarpent: In fact, with EigenPy, you can return an Eigen::Matrix as either numpy.Matrix or numpy.Array depending on the convention you want to choose. By default, we use the notion of Matrix.

rhaschke: I'm not sure that this should be a _global_ flag.

jcarpent: The class is a Singleton, so it can be.

PR #66, opened 2019-06-28
core: make np_type static thhrough a Singleton

jcarpent:
Fix a bug related to duplicated objects when linking.


include/eigenpy/details.hpp
```diff
@@ -89,18 +89,19 @@ namespace eigenpy
...
-      getInstance().np_type = MATRIX_TYPE;
+      getType() = MATRIX_TYPE;
     }
     
-    static NP_TYPE getType()
+    static NP_TYPE & getType()
     {
-      return getInstance().np_type;
+      static NP_TYPE np_type;
+      return np_type;
```
jmirabel: Are you sure you solved the issue ?
I think you should move the definition of this function to a non-installed cc file.

jcarpent: Both are working of course.

Issue #68, opened 2019-07-11
Why are Python and NumPy system include paths?


wxmerkt:
From MoveIt (https://github.com/ros-planning/moveit/issues/1536):

> Instead upstream relies on pkg-config, but incidentally their pkgconfig file is broken too - it uses -isystem to include numpy which pkg-config --cflags-only-I eigenpy fails to extract.

Indeed, they are currently set as system includes:
https://github.com/stack-of-tasks/eigenpy/blob/aacf9982064140368ebacfcece40db5ed33e3d76/CMakeLists.txt#L53-L55

And when running `pkg-config --cflags-only-I eigenpy`, indeed the NumPy is missing. They do show up on `pkg-config --cflags-only-other eigenpy` though, as expected. 

Is there a specific reason to set them as `SYSTEM`?

Edit: From [CMake docs](https://cmake.org/cmake/help/v3.0/command/include_directories.html):
> If the SYSTEM option is given, the compiler will be told the directories are meant as system include directories on some platforms. Signalling this setting might achieve effects such as the compiler skipping warnings, or these fixed-install system files not being considered in dependency calculations - see compiler docs.

jcarpent:
The main reason is to avoid warnings coming from dependencies when compiling.

jcarpent:
I think I can close this issue. @wxmerkt Feel free to reopen it if needed.

PR #69, opened 2019-07-11
Sync CMake submodule, make Python non-system include directories, add package.xml

wxmerkt:
This PR is a proposal to include a few changes to make eigenpy build as a 3rd party project in a catkin workspace and get rid off the catkin wrapper crutch:

- Update CMake submodule to include the python-install-path fix
- Change NumPy and Python to not use SYSTEM include directories and also change them in the pkg-config export (to address #68). This may be a significant change and I am not sure of the reasons for the initial selection - could you please help me understand it @jcarpent @nim65s? [And whether it could have possible downstream issues?] If it does, what else can we do to change, e.g., leave as SYSTEM include but change in pkg-config export? Or would that be a problem as well?
- Add a package.xml to build as external CMake project (to make eigenpy_catkin as a wrapper redundant)


CMakeLists.txt
```diff
@@ -50,9 +50,9 @@ OPTION (BUILD_UNIT_TESTS "Build the unitary tests" ON)
 
 FINDPYTHON()
 
-INCLUDE_DIRECTORIES(SYSTEM ${PYTHON_INCLUDE_DIRS})
+INCLUDE_DIRECTORIES(${PYTHON_INCLUDE_DIRS})
```
wxmerkt: Thinking about it more, I possibly should revert the change to Python for sure, but may leave for NumPy?

rhaschke:
While I think using non-system includes is the correct approach, I'm not sure the genuine `eigenpy` project should be cluttered with a package.xml for ROS.

v4hn:
In theory it should not matter whether or not the package itself builds with `SYSTEM` flags or not.
The important part is that they really should not *export* includes as `SYSTEM` includes.
Downstream dependencies have to decide whether they want it that way or not, not the library.

Adding a `package.xml` file is the prime example for using the release repository.
However, if the package's maintainers allow that file upstream (and maintain the version number in there), this simplifies things.

Last but not least, this request fixes the package-config file, which is a plain bug fix, but it would be even better to generate a `eigenpyConfig.cmake` file too. The package already uses a huge cmake infrastructure which provides macros to create these.

CMakeLists.txt
jcarpent: You should indeed keep it here, and you can remove it for the pkgconfig part.

jcarpent:
@v4hn @rhaschke Thanks you for taking the time to provide feedback.

@v4hn We will try to generate the eigenpyConfig.cmake along with the pkgconfig file.

jcarpent:
@wxmerkt It appears that adding package.xml inside the main directory of `eigenpy` makes travis failing. 
What is the advantage of having `package.xml` inside this repository or in another ros package repository?

I can manage this error by changing the behavior of the submodule travis contained in `eigenpy`. What do you think is better?

wxmerkt:
I don't understand why it makes travis fail - is there a particular reason it is picking up on it? I'd like to learn more why this happens, as my new plan would be to replace the ExternalProject-catkin wrappers all around with just adding a package.xml to the repositories and then using the JRL CMake. End users can use whatever they would before but it builds just fine in a catkin workspace (tested locally with pinocchio, hpp-fcl, eigenpy, tsid - all without modifications to upstream apart from adding that package.xml).

The advantage is people can directly clone this repository into their workspace (having it in a release repository means people need to clone and manually patch - release is only for release to buildfarm, not local builds). A good example is e.g. OctoMap.

jcarpent:
#71 has been merged. I will launch again Travis.

jcarpent:
It seems #71 has solved the compilation issue on Travis.
Thanks @wxmerkt for your nice contribution.

wxmerkt:
Thank you very much for fixing Travis and all the support :-) How can we generate the `eigenpyConfig.cmake` with the cmake submodule? I saw `project_finalize_package_config()` but couldn't get it to work proper (I tested it with pinocchio).

PR #92, opened 2019-10-15
Expose all real-valued specialised Eigen types

wxmerkt:
We are using some of the specialised Eigen types (`RowVectorXd`, `Matrix2Xd`). Without exposing them, EigenPy-dependent projects throw that the incorrect signature was used. This PR exposes all real-valued types.


jcarpent:
@wxmerkt By the way, do you know that for each project, you can expose the related types which are specific to your project as done here with Pinocchio:
https://github.com/stack-of-tasks/pinocchio/blob/380fdefb172efa07f2dd9112ce2d389e25aa7a41/bindings/python/module.cpp#L38
?

wxmerkt:
Thank you for merging. Can we make a robotpkg release with these changes?

nim65s:
Sure. If nobody is against, I'll make a 1.6.1 tonight.

Issue #94, opened 2019-10-21
quaternion.toRotationMatrix() does not work


gabrielebndn:
These lines
```
import eigenpy
q = eigenpy.Quaternion(1.,0.,0.,0.)
q.toRotationMatrix()
```
result in the following error
```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-3-676a92d9f600> in <module>()
----> 1 q.toRotationMatrix()

AttributeError: 'Quaternion' object has no attribute 'toRotationMatrix'
```

I do not understand this, as I can clearly see in https://github.com/gabrielebndn/eigenpy/blob/72f57d877e2b9fb7b0d2d216ce947ad62073286f/include/eigenpy/quaternion.hpp#L106 that `toRotationMatrix` is exposed, correctly, it seems.
What is weirder is that `ipython` actually *suggests* `toRotationMatrix` as an available method and auto-completes it, but then it gives an error as if it did not know it if you actually try to use it. What is going on?

This is not super-important as the same functionality is provided by the method `matrix()`. The duplication comes from Eigen itself. The methods are inherited from  [RotationBase](https://eigen.tuxfamily.org/dox/classEigen_1_1RotationBase.html#title14), where the documentation seems to suggest that the preferred method should be `toRotationMatrix`, while `matrix` was only "added to be conform with the Transform class' naming scheme." Further, in [QuaternionBase](https://eigen.tuxfamily.org/dox/classEigen_1_1QuaternionBase.html) only `toRotationMatrix` is documented

jcarpent:
I've found the bug. I will open a PR fixing the issue.
@gabrielebndn Thanks for the useful feedback.

jcarpent:
Fixed by #95.

PR #95, opened 2019-10-21
quaternion: fix exposure of toRotationMatrix

jcarpent:


jcarpent:
Related to #94

include/eigenpy/quaternion.hpp
```diff
@@ -103,7 +103,7 @@ namespace eigenpy
       .def("coeffs",(const Vector4 & (Quaternion::*)()const)&Quaternion::coeffs,
            bp::return_value_policy<bp::copy_const_reference>())
       .def("matrix",&Quaternion::matrix,"Returns an equivalent rotation matrix")
-      .def("toRotationMatrix ",&Quaternion::toRotationMatrix,"Returns an equivalent 3x3 rotation matrix.")
+      .def("toRotationMatrix",&Quaternion::toRotationMatrix,"Returns an equivalent 3x3 rotation matrix.")
```
gabrielebndn: Can't believe this ^^' I hadn't noticed! All because of one extra space! Noce one!

gabrielebndn:
Surprising boost-python even allows this!
How about adding `toRotationMatrix()` in the unit tests?
Also, a little explanation in the doc to say that `matrix()` is the same as `toRotationMatrix()`?

Issue #98, opened 2019-10-23
Generate and install CMake config


wxmerkt:
Following https://github.com/jrl-umi3218/jrl-cmakemodules/issues/247, can we please add the necessary features to generate and install the cmake config such that the only requirements to use eigenpy reduce to:

```cmake
find_package(eigenpy REQUIRED)
```

And can we please tag a release using the latest two PRs and this CMake configuration.

(I tried it following the comment in https://github.com/jrl-umi3218/jrl-cmakemodules/issues/247, but didn't get it to install yet)

jcarpent:
Solved by #132.

PR #99, opened 2019-10-23
Update package.xml to fix unstable build warnings

wxmerkt:
This PR fixes the unstable build warnings we are getting via email every time a PR is merged to master. It also updates the version number to the yet-to-be-released 1.6.2 which I suggest to tag and release once this PR is merged.

After this PR, I will make a new ROS buildfarm release which also changes from the previous catkinized version to regular third-party library. As part of this, I will update MoveIt to comply with the pkg-config search (as I believe #98 may take longer).

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0"?>
 <package format="2">
   <name>eigenpy</name>
-  <version>1.5.3</version>
+  <version>1.6.2</version>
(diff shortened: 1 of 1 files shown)


jcarpent:
Yes, I think #98 will take some time. Hopefully next week I may give a try.

wxmerkt:
Thank you for merging this quickly. Can you please tag version 1.6.2?

PR #107, opened 2019-11-07
Update Quaternion class with more constructors

jcarpent:


include/eigenpy/quaternion.hpp
```diff
@@ -65,12 +65,13 @@ namespace eigenpy
     {
       cl
       .def(bp::init<>("Default constructor"))
-      .def(bp::init<Matrix3>((bp::arg("matrixRotation")),"Initialize from rotation matrix."))
-      .def(bp::init<AngleAxis>((bp::arg("angleaxis")),"Initialize from angle axis."))
-      .def(bp::init<Quaternion>((bp::arg("clone")),"Copy constructor."))
+      .def(bp::init<Vector4>((bp::arg("Vec4: a 4D vector representing quaternion coefficients")),"Initialize from a vector 4D."))
```
wxmerkt: It may be a good idea to include the convention in the help message (wxyz or xyzw) in case someone didn't double check the Eigen documentation

jcarpent: @wxmerkt You're totally right.

PR #109, opened 2019-11-08
eigenpyConfig.cmake: find absolute library paths for all required libs

rhaschke:
As cmake expects absolute library paths, while pkgconfig provides lib names only (together with
LIBRARY_DIRS), one needs to transform names to full library paths.

This fixes 7854c0f / #101. Without this fix, the MoveIt catkin build still fails with:
```
Errors     << moveit_ros_planning_interface:make /vol/sandbox/ros/logs/moveit_ros_planning_interface/build.make.000.log
/usr/bin/ld: cannot find -leigenpy
collect2: error: ld returned 1 exit status
```

### DIFF
--- a/eigenpyConfig.cmake
+++ b/eigenpyConfig.cmake
@@ -2,5 +2,13 @@ cmake_minimum_required(VERSION 2.8.3)
 
 message(STATUS "Loading eigenpy from PkgConfig")
 
-find_package(PkgConfig)
+find_package(PkgConfig REQUIRED)
 pkg_check_modules(eigenpy REQUIRED eigenpy)
+
+# find absolute library paths for all eigenpy_LIBRARIES
+set(libs ${eigenpy_LIBRARIES})
+set(eigenpy_LIBRARIES "")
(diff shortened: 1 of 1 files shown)


wxmerkt:
Thank you for the fix @rhaschke. I think this is neater than #108, which also works in compiling MoveIt.

rhaschke:
Yeah, using `link_directories` is "bad" style, because it might introduce unintended side effects.

rhaschke:
@wxmerkt, maybe you want to close #108 in favor of this PR?

rhaschke:
Thanks for this quick merge! 
@wxmerkt, can you please prepare a new (Kinetic and Melodic) release to fix the issues on the ROS build farm? Thanks.

jcarpent:
I will handle it ;)

wxmerkt:
Kinetic: https://github.com/ros/rosdistro/pull/22934
Melodic: https://github.com/ros/rosdistro/pull/22935

PR #113, opened 2019-11-13
Corrected PYTHON_EXT_SUFFIX usage.

seanyen:
`${PYTHON_EXT_SUFFIX}` (`.so` for `Linux` and `MacOS X`, and `.pyd` for `Windows`) should be applied to Python extension module instead of the native shared library.


jcarpent:
I was also surprised at first glance. Thanks for the fix.

Issue #116, opened 2019-11-14
ROS Kinetic regression (lib install includes architecture)


wxmerkt:
This issue is to track the issues with the `eigenpy` package distributed via the ROS buildfarm on Ubuntu 16.04 which affects `pinocchio` (regression, does not build) and `moveit` (works because added `eigenpyConfig.cmake`). Currently, only `pinocchio` and other **pkg-config consumers** are broken.

**Note: This problem only happens on 16.04/kinetic due to the old version of `debhelper`.**

- eigenpy uses `GNUInstallDirs`. On the `ros_buildfarm`, the `CMAKE_INSTALL_FULL_LIBDIR` includes the system architecture for the `pkg-config` file, while the library is correctly installed in `lib/`.
- `libdir` and `pkglibdir` both include the system architecture:

```
cat /opt/ros/kinetic/lib/x86_64-linux-gnu/pkgconfig/eigenpy.pc 
prefix=/opt/ros/kinetic
exec_prefix=/opt/ros/kinetic
**libdir=/opt/ros/kinetic/lib/x86_64-linux-gnu**
bindir=/opt/ros/kinetic/bin
**pkglibdir=/opt/ros/kinetic/lib/x86_64-linux-gnu/eigenpy**
includedir=/opt/ros/kinetic/include
datarootdir=/opt/ros/kinetic/share
pkgdatarootdir=/opt/ros/kinetic/share
docdir=/opt/ros/kinetic/share/doc/eigenpy
doxygendocdir=/opt/ros/kinetic/share/doc/eigenpy/doxygen-html
```

Related issues:
- Issue identified here: https://github.com/ros-infrastructure/ros_buildfarm/issues/543#issuecomment-395249636
- https://github.com/ros-infrastructure/ros_buildfarm/issues/686
- https://github.com/ros/catkin/issues/545

@tfoote: Is there any way of resolving this apart from changing the `rules` in the packaging manually (e.g. https://github.com/mrpt-ros-pkg-release/mrpt1-release/commit/17361d75ef9e73c3496dca74813c4b9ee90ec643)? The most relevant answer to the issue here is from https://github.com/ros-infrastructure/ros_buildfarm/issues/543#issuecomment-395249636

tfoote:
There have been some other projects that hit this bug in the past too: https://github.com/ros/console_bridge/issues/15

With a follow up here: https://github.com/ros/console_bridge/pull/19

wxmerkt:
[...]

PR #120, opened 2019-11-15
Fix install directory, address #116

wxmerkt:
The issue underlying #116 is that:

1. The library gets installed to `${CMAKE_INSTALL_PREFIX}/lib`
2. However, the use of `GNUInstallDirs` inside the `cmake/base.cmake` sets the variable defining what gets set in the `pkg-config` template to use the `FULL` variables. These get set to include the prefix on some platforms (and old versions of debhelper). 
3. As a result there is a mismatch between what the `eigenpy.pc` says for where the library and pkg-config files are and where they actually get installed.

This took quite some time to debug...


wxmerkt:
@nim65s Will this be fine for you guys? I reckon this issue may lurk in other cmakemodules depending projects as well...

jcarpent:
@wxmerkt Thanks a lot for fixing this annoying bug.

nim65s:
If it fixes the build for you, and if that's the right way to do it, then it's ok for me. 
For robotpkg, I don't think it will change something, and if it does, I will be able to override it anyway :)

nim65s:
(i would have preferred `lib` to `${CMAKE_INSTALL_FULL_INCLUDEDIR}`, but this is not relevant for now.)

jcarpent:
@nim65s Thanks for your feedback.

wxmerkt:
Thank you guys! @jcarpent Can you please increment the package.xml and make a new release? :-)

PR #121, opened 2019-11-15
cmake: sync submodule

jcarpent:


jcarpent:
@wxmerkt I will do the release now.

PR #128, opened 2019-11-25
Towards a full CMake export of the project

jcarpent:


.gitignore
```diff
@@ -1,6 +1 @@
-build*
```
wxmerkt: What is the benefit of removing `build` from the ignore list? If someone follows the regular workflow, they would create either `build/` or `_build/`. Having it on the gitignore ignores it from accidental commits and allows `git clean -fdx` to reset hard (beyond `git reset --hard`)

jcarpent: Sure. My bad. Thanks for the comment.

PR #129, opened 2019-12-04
python: install lib inside site-packages/eigenpy

jcarpent:


python/eigenpy/__init__.py
```diff
@@ -14,5 +14,4 @@
 # eigenpy If not, see
 # <http://www.gnu.org/licenses/>.
 
-import numpy as np
 from eigenpy import *
```
jmirabel: shouldn't it be `from .eigenpy import *` ?

jcarpent: Yes, for Python3 you're right.

PR #132, opened 2019-12-05
Improve packaging and export of the project

jcarpent:
@wxmerkt I have implemented the full export of the project. It should work now. Could you check that the export is working well?

Before merging and making a new release, I will implement a small unit test to check the export is working well too.


wxmerkt:
Thank you for the great effort. I will check it either tomorrow or Monday (and also check that it doesn't break MoveIt)

jcarpent:
@wxmerkt I've added a unit test of using the new export. It seems to work well.

CMakeLists.txt
```diff
@@ -9,6 +9,7 @@ SET(PROJECT_NAME eigenpy)
 SET(PROJECT_DESCRIPTION "Bindings between Numpy and Eigen using Boost.Python")
 SET(PROJECT_URL "http://github.com/stack-of-tasks/eigenpy")
 SET(PROJECT_USE_CMAKE_EXPORT TRUE)
+SET(PROJECT_USE_KEYWORD_LINK_LIBRARIES TRUE)
```
wxmerkt: What does this do?

wxmerkt: Nevermind, just found it.

jcarpent:
@wxmerkt A kind reminder. If you have time today to check this PR ;)

wxmerkt:
I have just tested it and it appears to work.

However, please note: The install path for the `eigenpyConfig.cmake` is changing from `share/cmake/eigenpy` to `lib/cmake/eigenpy` (!!).

jcarpent:
> The install path for the `eigenpyConfig.cmake` is changing from `share/cmake/eigenpy` to `lib/cmake/eigenpy` (!!).

Is it an issue?

wxmerkt:
It appears to work - we should note it in case someone comes back in future if it breaks something. From me, this PR is good to go

jcarpent:
Do you want me to install it in the share folder too?

wxmerkt:
No, let's leave it as is (no duplication) until we have a use case that requires it

jcarpent:
@wxmerkt Thanks a lot for your nice feedback.

PR #134, opened 2019-12-07
[CI] use memmos instead of eur0c

nim65s:
https://gepgitlab.laas.fr/gsaurel/eigenpy/pipelines/7470


jcarpent:
@nim65s Thanks a lot for this fix. Do you think that we can also add a hooker towards the gitlab-ci?

nim65s:
@jcarpent : add a hook is still in my todolist.

We can either use the hook provided by Gitlab, in which case we would have to pay ~10k€/month to get the version with this functionnality, or write our own.

Writing one is in my todo list for January.

Issue #87, opened 2019-09-19
Timeline for using np.array by default.


jmirabel:
Let's open the discussion to decide a timeline for using numpy arrays by default, instead of numpy matrices.

cmastalli:
I vote for the beginning of the next year. So we give a couple of month to everyone adapt their code.

nim65s:
Does this change in defaults settings means that we have anything to adapt ?

jmirabel:
Arrays always do coefficient-wise operations while matrices do not. The most annoying difference is for the operator \*.

nim65s:
ok, thanks

jmirabel:
I think `Array @ Array` does the same thing as `Matrix * Matrix`.

For a smooth change, the best idea I have is to echo a warning the first time an object of type Matrix is returned, only if `eigenpy.switchToNumpyMatrix()` has not been called explicitly.

It is easy to implement and it forces people using matrices to explicitly say they want matrices. 3 months after the release of this warning, I suggest to release a new version of EigenPy, increasing the major number.

nmansard:
The problem of Array@Array is that it breaks the compatibility with 
Python 2.

I agree with your last comment: 3 months after the release of the 
warning, we can change the default behavior of EigenPy (knowing that 
people that are late can keep the previous behavior 
|eigenpy.switchToNumpyMatrix()|).

jcarpent:
Concerning Python2, we won't support it after January 1, 2020 (https://pythonclock.org/).

proyan:
> I vote for the beginning of the next year. So we give a couple of month to everyone adapt their code.

If we want people to adapt their code, the warnings should be issued from now. Then from January, we can move to np.array completely.

jmirabel:
@proyan that's what I suggested.

jcarpent:
Solved by #89. I will provide a new release soon.

nim65s:
I'll wait for this new source release, and after that I will launch the binary releases (eigenpy & pinocchio) on robotpkg

jmirabel:
Time is up !

jmirabel:
Should we change the default ?

wxmerkt:
Happy to change the default for array/matrix. However, we should not change the default for Python2/3 as Python2 will be supported on LTS platforms until they reach their EOL.

gabrielebndn:
I would say that we should wait at least until https://github.com/stack-of-tasks/pinocchio/issues/1026 is solved

[...]

PR #150, opened 2020-01-30
Switch to np.array type by default

jcarpent:
Solving #87.

### DIFF
--- a/include/eigenpy/details.hpp
+++ b/include/eigenpy/details.hpp
@@ -155,7 +155,6 @@ namespace eigenpy
 
   enum NP_TYPE
   {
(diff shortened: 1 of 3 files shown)


gabrielebndn:
You are aware Pinocchio isn't 100% ready?
I can fix it in one or two days

jcarpent:
I will proceed to a new release today or one the following days.

gabrielebndn:
May I suggest increasing the major version for next release? After all, it is a big change which will likely break lots of stuff for users who are not ready

jmirabel:
I think this was what we decided in #87

PR #166, opened 2020-02-27
Add Windows CI

jcarpent:


.github/workflows/build.yml
```diff
@@ -0,0 +1,48 @@
+name: Build Eigenpy
+on:
+  pull_request:
+  push:
+    branches:
+      - master
```
seanyen: I forgot the `devel` branch. You can enable both.

jcarpent: You're right. I will do.

PR #170, opened 2020-02-28
[CI] update gitlab CI

nim65s:
Hi,

This PR updates the gitlab CI configuration from the generated configuration at https://rainboard.laas.fr/project/eigenpy/.gitlab-ci.yml (which is now accessible from the LAAS, or with LAAS credentials, or with a custom account)

This:
- add the doc-coverage target, to save the generated documentation an code-coverage results
- add the format target, to check the code formatting (allow_failure is activated here for now), ref https://github.com/gepetto/linters
- set 18.04 as the main platform (for now, this affects only format & doc-coverage target)
- allow automatic retries in case of runner failures
- drop 14.04 checks

Before: https://gepgitlab.laas.fr/stack-of-tasks/eigenpy/pipelines/8522
After: https://gepgitlab.laas.fr/gsaurel/eigenpy/pipelines/8523

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,6 +1,6 @@
+# https://rainboard.laas.fr/project/eigenpy/.gitlab-ci.yml
+
 variables:
-  GIT_SUBMODULE_STRATEGY: "recursive"
-  GIT_DEPTH: "3"
   CCACHE_BASEDIR: "${CI_PROJECT_DIR}"
   CCACHE_DIR: "${CI_PROJECT_DIR}/ccache"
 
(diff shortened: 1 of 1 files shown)


jmirabel:
Do you think you can make gitlab use ccache ?

nim65s:
gitlab already uses ccache I think, I'll check.

nim65s:
I checked, and I think that ccache is correctly configured in this .gitlab-ci.yml, and in the docker images using robotpkg's "ccache-gcc" as "prefered alternative" for compilers.

But I think this is not as helpful as it should be because the jobs are run on many different runners, themselves on many different computers, and I think that there is one cache per runner.

I have been looking for solutions to this for a long time:
- I have been following https://github.com/ccache/ccache/pull/58 for a long time in the hope of setting a central cache server for all the team's builds. but this is not coming soon. 
- I also looked at https://docs.bazel.build/versions/2.0.0/remote-caching.html, but this doesn't work with multiple compilers, and therefore will not work with our CI with multiple linux distributions.
- I also tried https://github.com/mozilla/sccache, but this wasn't easy to setup for all our uses cases at all.

If anyone as other / better solutions, this would be really cool… But in fact, I also hope that this will be fixed on the new runner infrastructure with kubernetes we are building at LAAS, as they will probably be really better at sharing their caches :)

jmirabel:
Maybe artifacts could help.

jcarpent:
I will make a new release, as the bug reported in https://github.com/jrl-umi3218/jrl-cmakemodules/issues/333 is now fixed.

nim65s:
@jmirabel : artifacts is a thing to test, yes, I'll put that in my todolist. But if someone gets results with that, please share :)

jmirabel:
I said that because I think it's what travis uses (e.g. see https://travis-ci.org/humanoid-path-planner/hpp-fcl/jobs/656176576).

nim65s:
From this, it looks like this is only about sharing the `/home/travis/.ccache` folder with other jobs, which is exactly what I'm doing with https://github.com/stack-of-tasks/eigenpy/blob/master/.gitlab-ci.yml#L9 and the `${CI_PROJECT_DIR}/ccache` directory.

PR #183, opened 2020-04-03
Improve packaging testing

jcarpent:


jcarpent:
@wxmerkt I will make the release just after this PR is merged.

PR #186, opened 2020-04-22
Fix LLT with current Eigen master branch

gleichdick:
Return type of LLT::rankUpdate was changed from LLT (value) to LLT& (reference) on Eigen master branch


gleichdick:
You are welcome.
No thanks, as I'm building from source with `-std=c++17` I don't need a new release

PR #187, opened 2020-04-23
package.xml: Prepare for Ubuntu 20.04 / Python3

wxmerkt:
This adds compatibility with Ubuntu 20.04 and distributions that default to Python 3.

I've tested this with a CI based on 20.04.

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0"?>
-<package format="2">
+<package format="3">
   <name>eigenpy</name>
   <version>2.3.1</version>
   <description>Bindings between Numpy and Eigen using Boost.Python</description>
--- a/package.xml
+++ b/package.xml
@@ -13,8 +13,10 @@
 
   <build_depend>git</build_depend>
   <build_depend>doxygen</build_depend>
-  <depend>python</depend>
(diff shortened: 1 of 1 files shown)


package.xml
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0"?>
-<package format="2">
+<package format="3">
```
nim65s, line 2: Is this going to work on older systems ?
It's not even on the wiki yet: http://wiki.ros.org/catkin/package.xml

wxmerkt: It's supported since `rosdep` version 0.15.0: https://github.com/ros-infrastructure/rosdep/blob/master/CHANGELOG.rst#0150-2019-01-24

Do you see a system that relies on an older `rosdep` version?

It's been finalised in REP0149: https://www.ros.org/reps/rep-0149.html#package-format-3
And the XSD is here: http://download.ros.org/schema/package_format3.xsd

wxmerkt: ROS Kinetic / Ubuntu Xenial has 0.18.0 and I believe that is the oldest setting are supporting.

nim65s: ok, thanks :)

PR #201, opened 2020-07-02
[CMake] Add SUFFIX_SO_VERSION

### LINKED ISSUE #200: on robotpkg, eigenpy library should have soversion
Maybe on robotpkg, eigenpy library should be suffixed with the version number. @nim65s can you handle it ?

For pinocchio, we may do the same thing although it isn't as problematic right now (but may become in the future).

_Originally posted by @jmirabel in https://github.com/stack-of-tasks/pinocchio/issues/1229#issuecomment-653026773_

nim65s:
fix #200


CMakeLists.txt
```diff
@@ -45,6 +45,7 @@ ENDIF(WIN32)
 # --- OPTIONS  ---------------------------------------
 # ----------------------------------------------------
 OPTION(INSTALL_DOCUMENTATION "Generate and install the documentation" FALSE)
+OPTION(SUFFIX_SO_VERSION "Suffix library name with its version" OFF)
```
jcarpent, line 48: We should decide a common naming for all the projects!

jmirabel:
Thank you @nim65s

CMakeLists.txt
nim65s: I agree. 

This is the name used in many stack-of-tasks projects (eg. [sot-core](https://github.com/stack-of-tasks/sot-core/blob/9109cd8c89130c22143b8ad54c5605b2c08b7b7c/CMakeLists.txt#L15), [dynamic-graph](https://github.com/stack-of-tasks/dynamic-graph/blob/d0431fb825b9049ee250c9c4c932f64f715e8ea4/CMakeLists.txt#L11)). 

I think this option under this name was added by @florent-lamiraux in https://github.com/stack-of-tasks/dynamic-graph/pull/59, and other related projects got the same option

jcarpent: TAG_LIBRARY_VERSION is for instance used for Pinocchio and related projects.

Issue #202, opened 2020-07-06
Build error with numpy 1.19.0


Ozon2:
Hello, there is an error compiling eigenpy on Arch Linux with numpy 1.19.0.

```
/home/ozon/.cache/yay/eigenpy/src/eigenpy-2.4.0/include/eigenpy/ufunc.hpp:149:5: erreur: pas de correspondance pour convertir la fonction « binary_op_greater_equal » vers le type « PyUFuncGenericFunction » {aka « void (*)(char**, const long int*, const long int*, void*) »}
  149 |     EIGENPY_REGISTER_BINARY_UFUNC(greater_equal,code,Scalar,Scalar,bool);
      |     ^
/home/ozon/.cache/yay/eigenpy/src/eigenpy-2.4.0/include/eigenpy/ufunc.hpp:48:5: note: les candidats sont : « template<class T> void eigenpy::internal::binary_op_greater_equal(char**, npy_intp*, npy_intp*, void*) »
   48 |     EIGENPY_REGISTER_BINARY_OPERATOR(greater_equal,>=)
      |     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/ozon/.cache/yay/eigenpy/src/eigenpy-2.4.0/include/eigenpy/ufunc.hpp:48:5: note:                      « template<class T1, class T2, class R> void eigenpy::internal::binary_op_greater_equal(char**, npy_intp*, npy_intp*, void*) »
   48 |     EIGENPY_REGISTER_BINARY_OPERATOR(greater_equal,>=)
      |     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/ozon/.cache/yay/eigenpy/src/eigenpy-2.4.0/include/eigenpy/ufunc.hpp:150:5: erreur: pas de correspondance pour convertir la fonction « binary_op_less_equal » vers le type « PyUFuncGenericFunction » {aka « void (*)(char**, const long int*, const long int*, void*) »}
  150 |     EIGENPY_REGISTER_BINARY_UFUNC(less_equal,code,Scalar,Scalar,bool);
      |     ^
/home/ozon/.cache/yay/eigenpy/src/eigenpy-2.4.0/include/eigenpy/ufunc.hpp:47:5: note: les candidats sont : « template<class T> void eigenpy::internal::binary_op_less_equal(char**, npy_intp*, npy_intp*, void*) »
[...]

PR #203, opened 2020-07-15
Fix compilation issues with NumPy 1.19

jcarpent:
Related to #202.


nim65s:
Hi,

This is a minor remark, but I don't understand why you didn't just add `const` instead of creating this`EIGENPY_NPY_CONST_UFUNC_ARG`.

jcarpent:
Because it depends on the version of Numpy. The two function signatures won't be the same.

Issue #220, opened 2021-01-18
Quaternion's constructor does not recognize a 4-element numpy array


gabrielebndn:
The following test:
```py
import numpy as np
import eigenpy as eig

v = np.array([0.5, 0.5, -0.5, 0.5])
q = eig.Quaternion(v)
```
fails with the following error:
```
Traceback (most recent call last):
  File "test_eigenpy.py", line 5, in <module>
    q = eig.Quaternion(v)
RuntimeError: The number of rows does not fit with the matrix type.
```

On the other hand, the slightly modified test below, using a two-dimensional column array, passes without problems:
```py
import numpy as np
import eigenpy as eig

v = np.array([[0.5], [0.5], [-0.5], [0.5]])
q = eig.Quaternion(v)
```

gabrielebndn:
Tested on: Ubuntu 20.04, Python 3.8.5, IPython 7.13.0, numpy 1.19.1

gabrielebndn:
I couldn't manage to understand where the problem is, because it seems 3-element arrays work fine:
```py
In [6]: eig.Quaternion(np.array([1., 2., 3.]), np.array([4., 5., 6.]))                                                                  
Out[6]: (x,y,z,w) = -0.0459784  0.0919568 -0.0459784   0.993638
```

gabrielebndn:
I think it could be easily be fixed by using `VectorXd` as an input and checking for the size, but it looks like a deeper problem with `Vector4` itself

gabrielebndn:
Notice the constructor from a single 4-element `numpy.array` isn't checked in the unit tests

gabrielebndn:
I think I am starting to understand the problem: in C++ there exists a single constructor accepting `MatrixBase` as input, which 
> Constructs and initializes a quaternion from either:
>     a rotation matrix expression,
>     a 4D vector expression representing quaternion coefficients.

https://eigen.tuxfamily.org/dox/classEigen_1_1Quaternion.html#aba694e0bbafcf5554b77cfe8ce69c14a

and the array gets treated like a matrix. If I do not expose the constructor from a 3D matrix, the array one works

gabrielebndn:
I've tried working on it with no satisfactory results, I'll leave it to you. I think my comment above is a good start

jcarpent:
Who is the you that you mention in the previous message?

gabrielebndn:
> Who is the you that you mention in the previous message?

[...]

PR #221, opened 2021-01-19
Fix init call order

jcarpent:
Related to #220


gabrielebndn:
I would have never thought that the solution would be as simple as this! I wonder how portable and robust it is...

gabrielebndn:
I confirm it works on my PC. I've made a small PR on your devel branch to add a unit test

PR #223, opened 2021-03-05
fix unit-tests for numpy 1.20

nim65s:
Hi,

The `py-user-type`unit test  stopped working on Arch with numpy 1.20 (which requires python >= 3.7).

I didn't see anything related to safe casts or  `num_type` changes in the [changelog](https://github.com/numpy/numpy/releases/tag/v1.20.0), but switching from `numpy<1.20` to `numpy >=1.20` easily shows that the tests were working in 1.19.5 and are failing since 1.20.0.

This PR tries to address that. 
The first commit seems legit to me, but the second feels more like a hack. 
I think we shouldn't get that `dtype('O')` in the first place. I checked that `dotfunc` was called in 1.19.5, but isn't called anymore on 1.20.0, so maybe the issue is more at the registration step.

### DIFF
--- a/include/eigenpy/numpy.hpp
+++ b/include/eigenpy/numpy.hpp
@@ -22,10 +22,10 @@
 
 #if defined _WIN32 || defined __CYGWIN__
   #define EIGENPY_GET_PY_ARRAY_TYPE(array) \
-    call_PyArray_ObjectType(reinterpret_cast<PyObject *>(array), 0)
+    call_PyArray_MinScalarType(array)->type_num
 #else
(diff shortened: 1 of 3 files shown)


jcarpent (changes requested):
I totally don’t understand the change you made. At the same time, don’t change the API.

jcarpent:
@nim65s Any feedback on this fix?

nim65s:
I need to make a better version of this, but this require investigation time to understand why numpy 1.20 started to return `NPY_OBJECT` instead of the expected user defined type.

jcarpent:
Duplicate with #228.

PR #232, opened 2021-05-19
Fix #231

### LINKED ISSUE #231: Quaternion leads to segfault with -march=native
I have compiled eigenpy with clang-10 (Release mode) and the following flags: `-march=native -mfpmath=sse`. Then, the following code leads to a segfault:
```python
import eigenpy
eigenpy.Quaternion.Identity()
```

cmastalli:
For some reason that I still don't understand, performing `.transpose()` in `coeffs()` is not compliance with `-march=native`.

This PR fixes #231.

### DIFF
--- a/include/eigenpy/quaternion.hpp
+++ b/include/eigenpy/quaternion.hpp
@@ -288,7 +288,7 @@ namespace eigenpy
     static std::string print(const Quaternion & self)
     {
       std::stringstream ss;
-      ss << "(x,y,z,w) = " << self.coeffs().transpose() << std::endl;
+      ss << "(x,y,z,w) = " << self.coeffs()[0] << " " << self.coeffs()[1] << " " << self.coeffs()[2] << " " << self.coeffs()[3] << std::endl;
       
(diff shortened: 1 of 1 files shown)


jcarpent:
This is not the correct way of solving your issue. Please see #233.

cmastalli:
> This is not the correct way of solving your issue. Please see #233.

Thanks @jcarpent for providing a solution. I have tested the other PR, and I confirm that it solved my issue.
Perhaps, it is still better the solution of provided in this PR as we don't need to manage a new object (simpler and effective solution).
However, I'll let you to decide.

jcarpent:
Your solution will not solve the problem. The problem is that the function was not returning an aligned Quaternion forced by Eigen. My PR solves this issue.

The error is in fact due to Boost.Python.

cmastalli:
> Your solution will not solve the problem. The problem is that the function was not returning an aligned Quaternion forced by Eigen. My PR solves this issue.

Thanks for the clarification :)

Issue #230, opened 2021-05-05
Import error


mauku:
Hi , 
i have installed the package for python 3.6 with the following command : sudo apt install robotpkg-py36-eigenpy
However , when I try to import eigenpy i get the following error : 
 import eigenpy
  File "/opt/openrobots/lib/python3.6/site-packages/eigenpy/__init__.py", line 17, in <module>
    from .eigenpy import *
ImportError: /usr/lib/x86_64-linux-gnu/libboost_python-py27.so.1.65.1: undefined symbol: PyClass_Type

It seems its compiled with boost lib for python 2.7 , any suggestions how to solve this ? 
Thanks in advance , Markus !

jcarpent:
Dear @mauku,

You should uninstall eigenpy for Python 2.7.
```
apt-get uninstall robotpkg-py36-eigenpy
```

On a large level, I would recommand to use conda to install eigenpy for easy and robust installation, if you are not confident with your current environment settings.

mauku:
Thanks for the fast answer. I already installed eigenpy for Python 2.7 . The issue stays the same. I would prefer the debian package over conda, since I don't have the best experiences with conda.

jcarpent:
In fact, you should uninstall this package. There is a conflict between eigenpy for Python 2.7 and eigenpy for Python 3.6

mauku:
Sorry i misswrote , i uninstalled the package , but the error stays

jcarpent:
Could you do: `locate libeigen.so`
And also `export LD_LIBRARY_PATH=/opt/openrobots/lib:$LD_LIBRARY_PATH`

mauku:
Doesn't find any libeigen.so only libeigenpy.so. But I think I found the issue. There was one libeigenpy.so file from ros melodic eigenpy package for python2. After removing all eigenpy files for ros it seems to work.

jcarpent:
Yes, I meant `libeigenpy.so`

wxmerkt:
> There was one libeigenpy.so file from ros melodic eigenpy package for python2. After removing all eigenpy files for ros it seems to work.

It's important not to mix and match EigenPy from ROS and EigenPy from RobotPkg particularly across different Python versions. It can be worked around by sourcing / setting the paths in the right order but your best bet is to use conda since it will fix the paths for you when activating the environment. Another workaround to removing - since this will break Moveit - would be _not to source ROS by default_

mauku:
[...]

PR #238, opened 2021-06-04
[CMake] add PYTHON_EXT_SUFFIX back to the main library

nim65s:
Hi,

This PR re-apply https://github.com/stack-of-tasks/eigenpy/pull/57, because the main library is linked to boost python, and in case we have a setup with multiple python versions, we want to ensure that the python module will not try to link to a version of the main library linked to the wrong boost python library.

This is a reccuring issue for users who are using Python 3 on Ubuntu 18.04 or 20.04, while they also have ROS, which is stuck to a Python 2
version of eigenpy.
eg. https://github.com/stack-of-tasks/eigenpy/issues/230

This was removed in https://github.com/stack-of-tasks/eigenpy/pull/113.

Instead of deactivating this feature for everybody because this can raise an issue on Windows (I have no idea what this issue is), this PR reactivate this feature for everybody except Windows users.

@seanyen : as the author of #113, I guess you were trying to fix something, so can you confirm that I am not breaking anything for you here ?


nim65s:
This was deleted by mistake.

CMakeLists.txt
```diff
@@ -182,6 +182,13 @@ TARGET_INCLUDE_DIRECTORIES(${PROJECT_NAME}
   $<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}/include>
   $<INSTALL_INTERFACE:include>)
 
+IF(APPLE)
```
jcarpent, line 185: Why did you change here?

nim65s: I'm just reapplying this: https://github.com/stack-of-tasks/eigenpy/commit/8cd9fd9a5d667b267ffbf9918a92237d120fbe49

If this is useless, I can remove it.

PR #240, opened 2021-07-16
Fix contruction of np.array from custom types

jcarpent:


jcarpent:
@nim65s Could you check the reasons of the failure on gitlab? As it is passing on Ubuntu/conda, this should work too on the LAAS build farm.

nim65s:
Hi @jcarpent,

I think this is a numpy version issue.

For example, in 20.04, this will fail with the same error message as in gitlab CI:

```dockerfile
FROM ubuntu:20.04                                                                                                      

ENV DEBIAN_FRONTEND=noninteractive CTEST_OUTPUT_ON_FAILURE=true

RUN apt-get update -y \
 && apt-get install -y \
    build-essential \
    cmake \
    git \
    libboost-all-dev \
    libeigen3-dev \
    python-is-python3 \
    python3-numpy \
    python3-pip \
 && rm -rf /var/lib/apt/lists/*

#RUN python -m pip install --upgrade numpy
RUN git clone --recursive -j4 -b devel https://github.com/jcarpent/eigenpy.git
WORKDIR eigenpy/build
RUN cmake ..
RUN make -sj16
RUN make test
``` 

Uncommenting numpy upgrade from 1.17.4 (ubuntu) to 1.21.0 (PyPI) make the tests pass as in conda CI.

proyan:
@nim65s I tried this branch on my local installation, with Ubuntu20.04, and numpy version 1.21.1, but couldn't reproduce this error. Perhaps this might be a numpy bug fixed between 1.21.0 and 1.21.1 🤞🏽

jcarpent:
as mentioned by @nim65s, the issue is coming for numpy < 1.21

proyan:
For Ubuntu20.04, numpy installation by apt (1.17.4) to and PyPI (1.21.1) both seem to be supported (if the bug really was in 1.21.0). If so, do we still need to provide a patch for numpy version <1.21.1? I guess we could simply ask the user to update the installed version

jcarpent:
We should not ask people to update their NumPy version to get the latest one. 
It would be more robust to support any recent versions coming natively with Ubuntu for instance. 
So, we should provide a patch.

unittest/python/test_user_type.py
```diff
@@ -7,8 +7,8 @@
 def test(mat):
   mat[:] = mat.dtype.type(10.)
   mat_copy = mat.copy()
-  assert (mat == mat_copy).all()
-  assert not (mat != mat_copy).all()
+  assert((mat == mat_copy).all())
```
nim65s, line 10: `assert`(https://docs.python.org/3/reference/simple_stmts.html#the-assert-statement) is a statement, not a function: it doesn't need parenthesis

proyan: I removed this commit.

proyan:
The CI failure for 
```
mat[:] = ...
```
is a similar issue as https://github.com/numpy/numpy/issues/19553
The issue appears in earlier version of numpy (1.17.x), and while later versions have fixed this assignment by access, the earlier version still has this bug.

unittest/python/test_user_type.py
```diff
@@ -1,14 +1,22 @@
 import user_type
+import numpy as np
+#from packaging import version
 
 rows = 10
 cols = 20
 
 def test(mat):
-  mat.fill(mat.dtype.type(10.))
+  mat[:] = mat.dtype.type(10.)
```
proyan, line 9: This assignment would fail for numpy version 1.17.4 which is the standard in ubuntu18.04LTS.
The bug doesn't exist anymore in 1.21.0, but I haven't checked in which version between it was fixed.
We might have to warn users about the issue when they import a custom user type, and to avoid certain assignments.

proyan:
I think we can remove the array assignment from the unit-test, merge this PR, and create an issue with label `won't-fix` to keep track of the numpy bug.

jcarpent:
I'm almost done with some updates to allow casting between types. Is it fine for you?

proyan:
For the simple test of :

```python
import user_type                                                                                                                                                                                               
v = user_type.CustomDouble(1)                                                                                                                                                                                  
print("address of v:",hex(id(v)))                                                                                                                                                                              
```
the following is the backtrace:
```gdb
address of v: 0x7fffdc229170

Thread 1 "python" received signal SIGSEGV, Segmentation fault.
0x000000000053acc5 in gc_list_remove (node=0x7fffdc229160, node=0x7fffdc229160) at ../Modules/gcmodule.c:2056
2056	../Modules/gcmodule.c: No such file or directory.
(gdb) bt
#0  0x000000000053acc5 in gc_list_remove (node=0x7fffdc229160, node=0x7fffdc229160) at ../Modules/gcmodule.c:2056
#1  PyObject_GC_Del (op=0x7fffdc229170) at ../Modules/gcmodule.c:2056
#2  0x00000000004d3bc4 in subtype_dealloc (self=<CustomDouble at remote 0x7fffdc229170>) at ../Objects/typeobject.c:1192

```

It looks like PyGC_Head is undefined/ill-defined for the newly created objects. Perhaps it is unable to understand the size of the new type. As a result, when the gc tries to search for the memory address [here](https://github.com/python/cpython/blob/v3.8.10/Modules/gcmodule.c#L2054), it simply decreases the pointer reference ([here](https://github.com/python/cpython/blob/main/Modules/gcmodule.c#L72)) and that is invalid. 

This is with Python 3.8, numpy 1.21.1. It would change with different versions.

jcarpent:
Thanks for the report. 
It seems that we need to register again the type, restore with 65ed89c.
Could you check 65ed89c solve the issue on your side?

jcarpent:
Gitlab seems to be happy (fantastic).

PR #253, opened 2021-08-24
Move block to set OPTIONS before including base.cmake

wxmerkt:
Setting INSTALL_DOCUMENTATION OFF wasn't respected as it was overridden
in post-project.cmake which gets called upon setting PROJECT(). Moving
the settings prior to the include of base.cmake, it works as intended.


jcarpent (approved):
Thanks for the fix @wxmerkt.

nim65s (approved):
thanks

PR #261, opened 2021-09-06
Remove useless setup

jcarpent:


jcarpent:
This PR also fixes bugs with Windows when converting from double to np.int32 and vice-versa

PR #267, opened 2022-02-07
Fix quaternion initialisation segfault under -march=native

### LINKED ISSUE #231: Quaternion leads to segfault with -march=native

wxmerkt:
Relates to and resolves #231

### DIFF
--- a/include/eigenpy/quaternion.hpp
+++ b/include/eigenpy/quaternion.hpp
@@ -110,12 +110,16 @@ namespace eigenpy
     void visit(PyClass& cl) const 
     {
       cl
-      .def(bp::init<Matrix3>((bp::arg("self"),bp::arg("R")),
-                             "Initialize from rotation matrix.\n"
-                             "\tR : a rotation matrix 3x3.")[bp::return_value_policy<bp::return_by_value>()])
-      .def(bp::init<AngleAxis>((bp::arg("self"),bp::arg("aa")),
-                               "Initialize from an angle axis.\n"
-                               "\taa: angle axis object."))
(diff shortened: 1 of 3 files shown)


include/eigenpy/quaternion.hpp
```diff
@@ -263,24 +268,42 @@ namespace eigenpy
...
-    
-    static Quaternion* FromTwoVectors(const Vector3& u, const Vector3& v)
+
+    static Quaternion* FromCoefficients(Scalar w, Scalar x, Scalar y, Scalar z)
+    {
+      Quaternion* q(new Quaternion(w, x, y, z));
+      return q;
+    }
+
+    static Quaternion* FromAngleAxis(const AngleAxis& aa)
```
wxmerkt, line 278: Note, this won't compile with `Eigen::Ref<AngleAxis>`

jcarpent: Why do you want to rely on `Eigen::Ref<AngleAxis>`?

jcarpent (approved):
Seems good to me.

jcarpent (commented):
The copy constructor works well?

include/eigenpy/quaternion.hpp
wxmerkt: Right, it doesn't matter here. `const Eigen::Ref<const TYPE>&` is my first approach out of habit

cmastalli:
I confirm that it fixed the issue from my side.
Thanks @wxmerkt ;)

PR #270, opened 2022-02-21
Fixed quaternion normalized (issue #268)

### LINKED ISSUE #268: Normalized function in Pinocchio Quaternion leads to segfaul under -march=native flag
The following Python code leads to segfaul if you compile `eigenpy` and `pinocchio` with `-march=native` flag.
```python
import pinocchio
import numpy as np

R = np.random.rand(3, 3)

# This leads to segtaul
print(pinocchio.Quaternion(R).normalized().coeffs())

# Instead, this patches the issue
quat = pinocchio.Quaternion(R)
quat.normalize() # Note that this is not normalized()
print(quat.coeffs())
```

fyi @wxmerkt

cmastalli:
Hey @jcarpent 

Thanks for trying to fix #268.
However, the PR https://github.com/stack-of-tasks/eigenpy/pull/269 didn't fix it.
Instead, this PR fixes the issue.

### DIFF
--- a/include/eigenpy/quaternion.hpp
+++ b/include/eigenpy/quaternion.hpp
@@ -194,7 +194,7 @@ namespace eigenpy
            "Normalizes the quaternion *this.")
       .def("normalized",&Quaternion::normalized,
            bp::arg("self"),
-           "Returns a normalized copy of *this.")
+           "Returns a normalized copy of *this.", bp::return_self<>())
       .def("squaredNorm",&Quaternion::squaredNorm,
            bp::arg("self"),
            "Returns the squared norm of the quaternion's coefficients.")


jcarpent (changes requested):
This is not good. Please look at the documentation of Eigen. Normalized returns a copy. There is no pointer relation between the two quantities.

cmastalli:
> This is not good. Please look at the documentation of Eigen. Normalized returns a copy. There is no pointer relation between the two quantities.

Are you sure?
My understanding is "it doesn't return a new object. Instead, it returns a copy of `*this`".

jcarpent:
I really don't know what you don't understand by: `Returns
    a normalized copy of *this `.

jcarpent:
Superseded by #271.

PR #271, opened 2022-02-22
Fix memory issue with Quaternion::normalized

jcarpent:
Related to #268.


include/eigenpy/quaternion.hpp
```diff
@@ -185,16 +185,19 @@ namespace eigenpy
            "Returns the quaternion describing the inverse rotation.")
       .def("setIdentity",&Quaternion::setIdentity,
            bp::arg("self"),
-           "Set *this to the idendity rotation.",bp::return_self<>())
+           "Set *this to the idendity rotation.",
```
wxmerkt, line 188:
```suggestion
           "Set *this to the identity rotation.",
```

PR #280, opened 2022-04-26
numpy: avoid deprecated header

ManifoldFR:
The `numpy/noprefix.h` array is deprecated and should be removed (it can create problems with some libraries when using recent C++ standards)

- [x] checked tests still pass

Cross-reference: https://github.com/numpy/numpy/pull/12402
Same fix applied in: 
https://github.com/matplotlib/matplotlib/pull/12976


hrp2-14:
Hi ! This project doesn't usually accept pull requests on master.

PR #282, opened 2022-05-03
test/eigen_ref: test using Eigen::Ref as data member

ManifoldFR:
This commit adds a test where the user exposes a struct with an `Eigen::Ref` data member and shows how to expose it using the right `boost::python::return_value_policy`.


jcarpent:
@ManifoldFR Could you rebase on top of devel?

ManifoldFR:
> @ManifoldFR Could you rebase on top of devel?

Done!

jcarpent (approved):

PR #284, opened 2022-05-09
Fix for Refs to dynamic (1, N) blocks

ManifoldFR:
This PR:

* adds tests where user takes (1,n) or (n,1) blocks of matrices as `Eigen::Ref` and returns them to NumPy
* fixes the corner case where (n,1) was taken
* builds on fix for `RowMajor` in 471b8d65bff9240239e6d490100d97074a429952

### DIFF
--- a/include/eigenpy/numpy-allocator.hpp
+++ b/include/eigenpy/numpy-allocator.hpp
@@ -70,8 +70,7 @@ struct NumpyAllocator<Eigen::Ref<MatType, Options, Stride> > {
 
     if (NumpyType::sharedMemory()) {
       const int Scalar_type_code = Register::getTypeCode<Scalar>();
-      const npy_int R = (npy_int)mat.rows();
-      const bool reverse_strides = MatType::IsRowMajor || (R == 1);
+      const bool reverse_strides = MatType::IsRowMajor || (mat.rows() == 1);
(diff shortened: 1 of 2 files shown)


include/eigenpy/numpy-allocator.hpp
```diff
@@ -70,11 +70,12 @@ struct NumpyAllocator<Eigen::Ref<MatType, Options, Stride> > {
 
     if (NumpyType::sharedMemory()) {
       const int Scalar_type_code = Register::getTypeCode<Scalar>();
-
-      Eigen::DenseIndex inner_stride = MatType::IsRowMajor ? mat.outerStride()
-                                                           : mat.innerStride(),
-                        outer_stride = MatType::IsRowMajor ? mat.innerStride()
-                                                           : mat.outerStride();
+      const npy_int R = (npy_int)mat.rows();
+      const bool reverse_strides = MatType::IsRowMajor || (R == 1);
```
jcarpent, line 74:
```suggestion
      const bool reverse_strides = MatType::IsRowMajor || (mat.rows() == 1);
```

jcarpent (changes requested):
One quick minor change ;)

ManifoldFR:
Writing a commit for the const<EigenRef> case now

jcarpent (approved):

PR #291, opened 2022-07-14
cmake: relocatable package for recent CMake versions

nim65s:
Hi,

This enable some recent CMake features when they are available:
1. stop linking to libpython.so as this is not necessary. And actually, it's not even a problem if this lib is not available on the build or run system. Plus it get stripped anyway when `--as-needed` is activated.
2. provide only references to cmake targets instead of hardcoded absolute path in exported modules.

Enabling 2. allows to move around the installed project and / or its dependencies without issues, which is a required thing when we build / install on a dynamic location to ensure a proper isolation. 

This PR shouldn't change anything for older systems.


CMakeLists.txt
```diff
@@ -200,13 +215,23 @@ SET(${PROJECT_NAME}_SOURCES
...
   $<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}/include>
   $<INSTALL_INTERFACE:include>)
 
+if(TARGET Eigen3::Eigen)
+  TARGET_LINK_LIBRARIES(${PROJECT_NAME} PUBLIC Eigen3::Eigen)
+else()
+  TARGET_INCLUDE_DIRECTORIES(${PROJECT_NAME} SYSTEM PUBLIC ${EIGEN3_INCLUDE_DIR})
+endif()
+
+if(TARGET Python3::NumPy)
```
jcarpent, line 227: Maybe we should add an helper function in the cmake module to accound the linkage of a target to NumPy?

nim65s:
Ok, I'll move that into the JRL cmakemodule

jcarpent:
> Ok, I'll move that into the JRL cmakemodule

Perfect. I will simplify future adjustments.

nim65s:
CI looks OK, except for the job requiring https://github.com/jrl-umi3218/jrl-cmakemodules/pull/518 to be merged first

PR #294, opened 2022-07-20
Temporary fix for Python3::NumPy missing export

jcarpent:

PR #295, opened 2022-07-22
Fix python numpy linking error

wxmerkt:
Depends on https://github.com/jrl-umi3218/jrl-cmakemodules/pull/523

Alternative for #294


nim65s (approved):
Thanks !

PR #298, opened 2022-07-27
ci: test packaging

nim65s:
Same idea as #296, but for projects that don't have the jrl-cmakemodules.

I think the CI will fail here, let's check this :)

### DIFF
--- a/.github/workflows/macos-linux-conda.yml
+++ b/.github/workflows/macos-linux-conda.yml
@@ -53,6 +53,13 @@ jobs:
         make test
         make install
 
+    - name: Test packaging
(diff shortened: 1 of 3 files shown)


jcarpent (approved):

nim65s:
```
CMake Error at /usr/local/miniconda/envs/eigenpy/lib/cmake/eigenpy/eigenpyConfig.cmake:136 (FINDPYTHON):
  Unknown CMake command "FINDPYTHON".
Call Stack (most recent call first):
  CMakeLists.txt:4 (find_package)
```
Ok, let's fix this

PR #307, opened 2022-08-20
[CI] Always run PRERELEASE tests

wxmerkt:
Our current CI configuration only ran prerelease tests for ROS2. We recently had a CMake breakage for Python2

@nim65s Can you please look into why the new CMake does not correctly configure for Python2 on 18.04?

Cf.
https://github.com/ros/rosdistro/pull/34135#issuecomment-1221034543
https://github.com/stack-of-tasks/eigenpy-ros-release/issues/4

### DIFF
--- a/.github/workflows/ros_ci.yml
+++ b/.github/workflows/ros_ci.yml
@@ -11,14 +11,16 @@ jobs:
     strategy:
       matrix:
         env:
-          - {ROS_DISTRO: melodic}
-          - {ROS_DISTRO: noetic, BUILDER: catkin_tools}
+          - {ROS_DISTRO: melodic, PRERELEASE: false}
+          - {ROS_DISTRO: noetic}
           - {ROS_DISTRO: foxy}
           - {ROS_DISTRO: galactic}
-          - {ROS_DISTRO: foxy, PRERELEASE: true}
-          - {ROS_DISTRO: galactic, PRERELEASE: true}
+          - {ROS_DISTRO: rolling}
+          - {ROS_DISTRO: humble}
     env:
       CCACHE_DIR: /github/home/.ccache # Enable ccache
+      PRERELEASE: true
+      BUILDER: colcon
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2


nim65s:
Thanks for the report. Do you know if we can have access to the 2.7.11 deb somewhere to inspect it ? 
http://wiki.ros.org/Distributions/Staging is not super helping :/

wxmerkt:
Yes, we can - the way I do it is I pull the Docker image, then install nano to edit `/etc/apt/sources.list.d/ros1-latest.list` and change `/ros/` to `/ros-testing/`, apt update & apt install.

One thing I noticed is that the `find_package(Boost` does not have the `REQUIRED` keyword set because we aren't asking for it per the macro.

nim65s:
Ok, I think the issue is that we find python27, and moveit find python:

wxmerkt:
Thank you @nim65s! :-) Is there a reason why 2.7.10 was fine and the issue only appeared with the CMake update in 2.7.11?

nim65s:
Yes, in 2.7.10,  we were linking to `${Boost_PYTHON_LIBRARIES}` instead of modern `Boost::python[…]` targets. 
If we don't find a solution soon, we can simply revert that, as this would prevent relecotable packages only for python 2.

nim65s:
FTR, this is a CMake version issue. I can reproduce the issue by building eigenpy & moveit, but if I update CMake from 3.10 (default 18.04) to 3.24 (with pip), it works better. I'll try to see what is the difference between the 2 that break everything.

jcarpent:
Maybe we can use a FindBoost.cmake more advanced. See https://github.com/jrl-umi3218/jrl-cmakemodules/blob/f4e4be6a6f6577d86c9e2b09ae3a9bf1a545437e/boost.cmake#L40-L46 for potential bug.

wxmerkt:
I believe the CI issue here is unrelated to resolving the issue on ROS Melodic/18.04/Python2. Can we make a release with just the previous PR merged and I will sort out the tests here at a later stage? I'd like to get the other users of Pinocchio via ROS unblocked before figuring out what's going on with CI here

jcarpent:
> I believe the CI issue here is unrelated to resolving the issue on ROS Melodic/18.04/Python2. Can we make a release with just the previous PR merged and I will sort out the tests here at a later stage? I'd like to get the other users of Pinocchio via ROS unblocked before figuring out what's going on with CI here

Done

wxmerkt:
The CI failure now is from `py-user-type` segfaulting. I can reproduce that locally as well with some settings of LD_LIBRARY_PATH. Have you observed something like this and know of a workaround @nim65s?

nim65s:
I don't. All tests are passing correctly on all the packages I maintain. 
If you have an environment that can trigger this segfault, and another one that doesn't, can you compare them ?

jcarpent:
I guess Python 2.7 is the issue. Could we stop maintaining it?

nim65s:
I'd say we can schedule stopping its maintenance. April 2023 would be a good choice, as this is the end of the maintenance of 18.04, which still has python 2.7 by default.

wxmerkt:
Turned off the extensive tests for 18.04 and left them on for the other distros and it appears to have fixed it - we are good to go with this PR :-)

jcarpent (approved):
Thanks a lot for this work @wxmerkt

PR #312, opened 2022-08-25
relocatable eigenpy.pc

nim65s:
Hi,

When some downstream packages use pkg-config to use eigenpy, they are not relocatable anymore. Here is a tested fix for that.

The downside of removing this is that people using pkg-config to quickly compile something will need to add boost manually, but since boost doesn't provide any .pc file, we're doomed.


nim65s:
Alternatively, we can drop support for eigenpy to be used through pkg-config by other softwares that want to be relocatable.
CMake exports are the good solution anyway.

PR #313, opened 2022-08-26
add relocatable test

nim65s:
Hi,

This add a CI check to ensure we are relocatable. 

It basically check for occurrences of the source dir, the build dir, and the install dir absolute paths in the installed package.

The missing bit is looking for full paths of dependencies, but this would require installing all dependencies in a non standard prefix, so this will come later.


nim65s:
This is failing because the following install dirs full paths are in the installed project:
```
/RELOC/PFX/lib/cmake/eigenpy/eigenpyTargets-release.cmake:  IMPORTED_LOCATION_RELEASE "/RELOC/PFX/lib/libeigenpy.so"
/RELOC/PFX/lib/cmake/eigenpy/eigenpyTargets-release.cmake:list(APPEND _cmake_import_check_files_for_eigenpy::eigenpy "/RELOC/PFX/lib/libeigenpy.so" )
/RELOC/PFX/lib/cmake/eigenpy/eigenpyTargets.cmake:  INTERFACE_INCLUDE_DIRECTORIES "${_IMPORT_PREFIX}/include;/RELOC/PFX/include"
```

Let's fix this.

nim65s:
This is basically reverting #120 ; but I guess #116 is no longer relevant, as 16.04 is obsolete. Can you confirm, @wxmerkt ?

wxmerkt:
If the paths in the pkg-config and CMake config are fine, it should be good (the issue at the time was that jrl-cmakemodules used full dir paths). I believe with the use of `CMAKE_INSTALL_LIBDIR` this should work - unless there is another FULLDIR in jrl-cmakemodules that I didn't spot. I only spotted one of these just now - will this be an issue?

https://github.com/jrl-umi3218/jrl-cmakemodules/blob/9d6d417f45548728fc5be2ba66c36327766d0875/package-config.cmake#L212-L217

nim65s:
That one gets dynamically fixed thanks to `PATH_VARS`, ref. https://cmake.org/cmake/help/latest/module/CMakePackageConfigHelpers.html, so it's not actually an absolute path, but it gets expanded to one when another project read eigenpyConfig.cmake

jcarpent (approved):

PR #318, opened 2022-09-08
Fix flake8 errors

wxmerkt:
Should fix the failing format job on gepgitlab

### DIFF
--- a/benchmarks/bench-switch.py
+++ b/benchmarks/bench-switch.py
@@ -3,8 +3,8 @@ from __future__ import print_function
 import eigenpy
 import numpy as np
 
-import time
-import timeit
+import time  # noqa
+import timeit  # noqa
 
 from IPython import get_ipython
 
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,0 +1,2 @@
+[flake8]
+extend-ignore = E203
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,2 +1,3 @@
 [flake8]
 extend-ignore = E203
+max-line-length = 119
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,3 +1,3 @@
 [flake8]
 extend-ignore = E203
-max-line-length = 119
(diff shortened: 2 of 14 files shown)


benchmarks/bench-switch.py
```diff
@@ -3,8 +3,8 @@
 import eigenpy
 import numpy as np
 
-import time
-import timeit
+import time  # noqa
+import timeit  # noqa
```
nim65s, line 7: Couldn't we remove those imports ?

nim65s (approved):
Thanks for this !
I have a minor comment, and removed E203, as this is incompatible with black

setup.cfg
```diff
@@ -0,0 +1,2 @@
+[flake8]
+extend-ignore = E203
```
wxmerkt, line 2: The gepetto linter config has linelength=119 and exclude cmake/, which this file would omit. Should we add it here as well?

benchmarks/bench-switch.py
wxmerkt: When the benchmark is run with IPython it is being used but the linter cannot figure this out

setup.cfg
nim65s: Good catch, i'll add this extend ignore to gepetto linters. But the linelength will have to move to 88 for black :/

nim65s:
now yapf is complaining, but we should abandon yapf for black now :(

setup.cfg
nim65s: PS: excluding cmake is not mandatory anymore for flake, since https://github.com/jrl-umi3218/jrl-cmakemodules/pull/524

nim65s:
Actually, the format is already good for black, thanks to pre-commit, so we just need to tell that to gepetto-linters with a link in the readme :)

jcarpent (commented):
Thanks @wxmerkt for your contribution and thanks @nim65s for quick review.

PR #325, opened 2022-10-04
Copy std-vector and std-map from Pinocchio

### LINKED ISSUE #324: Proposal: port StdVectorPythonVisitor from Pinocchio's bindings
The header is here: https://github.com/stack-of-tasks/pinocchio/blob/devel/bindings/python/utils/std-vector.hpp (or a more recent version).
I recently suggested moving/copying the **`StdVectorPythonVisitor`** template from Pinocchio's bindings upstream to this repo, so that downstream packages don't have to import or copy the header from Pinocchio.

We could also import the **`StdContainerFromPythonList`** template.

What do we think of this ? I could do it if we choose to go ahead with it.

ManifoldFR:
I recovered the history using `git-filter-repo`, squashed and rebased it into a single big commit before merging the files' history into eigenpy.

**TODO**

* Include the `utils/copyable.hpp` header from Pinocchio
* Include the relevant tests

Closes #324


hrp2-14:

jcarpent (changes requested):
@ManifoldFR Thanks a lot for this PR.
Could you remove aligned_vector, which is not useful to have I do think?

ManifoldFR:
> @ManifoldFR Thanks a lot for this PR. Could you remove aligned_vector, which is not useful to have I do think?

Ok, will do so

ManifoldFR:
Rebased on devel.

fabinsch:
Now it should be fixed. I will manually squash my commits and force push.

jcarpent:
Nice work @ManifoldFR and @fabinsch. Thanks a lot.

PR #336, opened 2023-01-07
Add full support of vectorization

jcarpent:


jcarpent:
We can now use `-mavx2 -mavx512` without any problem.

PR #343, opened 2023-01-30
Fix issue with Boost.Python < 1.71

jcarpent:


include/eigenpy/fwd.hpp
```diff
@@ -81,6 +93,14 @@
 
 #include "eigenpy/expose.hpp"
 
+#ifdef EIGENPY_WITH_CXX11_SUPPORT
+#include <memory>
+#define EIGENPY_SHATED_PTR_HOLDER_TYPE(T) ::std::shared_ptr<T>
```
nim65s, line 98:
```suggestion
#define EIGENPY_SHARED_PTR_HOLDER_TYPE(T) ::std::shared_ptr<T>
```

jcarpent: Thanks for the quick feedback.

PR #344, opened 2023-02-01
Fix for Python 3.6

jcarpent:


include/eigenpy/quaternion.hpp
```diff
@@ -350,7 +350,7 @@ class QuaternionVisitor
 
  public:
   static void expose() {
-#if BOOST_VERSION / 100 % 1000 < 71
+#if PY_MAJOR_VERSION == 6 && PY_MINOR_VERSION == 2
```
nim65s, line 353: You're looking at python 6.2 instead of 3.6

jcarpent: You're right. I don't where my mind was ;)

PR #360, opened 2023-04-20
Add example showing how to bind virtual classes, passing to overrides…

ManifoldFR:
… by reference, registering shared_ptr, and overriding factories and data classes from Python


jcarpent (dismissed):

ManifoldFR:
CI on melodic failed due to type hints in the Python code it seems, maybe the Python version is (very) old. I'll fix it

nim65s:
melodic has python 2.7. Maybe we can drop this for eigenpy 3 :)

ManifoldFR:
OK everything works now.

PR #367, opened 2023-05-15
WIP: Expose boost::none_t type (and std::nullopt_t when available)

ManifoldFR:
This PR:

+ adds src/optional.cpp to expose none types
+ this allows to have None as a default arg value for functions (also, shows up in function signatures when stubs are enabled)
+ also allows to have none_t return/argument type

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -207,6 +207,7 @@ set(${PROJECT_NAME}_SOURCES
     src/quaternion.cpp
     src/geometry-conversion.cpp
     src/std-vector.cpp
+    src/optional.cpp
     src/version.cpp)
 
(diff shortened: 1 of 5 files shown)


jcarpent (commented):
None object is already exposed, no?

jcarpent (approved):

ManifoldFR:
> None object is already exposed, no?

Only when the C++ signature says `boost::optional<T>`. If for any reason there's a value that's explicitly `boost::none_t` Boost.Python will complain about a missing converter, for instance if you bind a function as follows:
```cpp
bp::def("myfunc", myfunc, (bp::arg("arg1"), bp::arg("arg2") = boost::none));
```

Actually this PR shouldn't have been merged yet, it was missing a check for the registration

jcarpent:
Ah, you should rather the feature of GitHub to mark a PR as not ready.

Issue #365, opened 2023-04-28
possible to use eigenpy with custom numeric scalar types?


ofloveandhate:
My package uses Boost.Multiprecision complex numbers.  Is it possible to extend eigenpy to include Eigen matrices / vectors / etc with custom scalar types?

jcarpent:
Good question and good feature of Eigenpy ;)

Yes, you can.
You just need to declare the type converters for that.
I've done that for [Pinocchio](https://github.com/stack-of-tasks/pinocchio).
Here is a code taken from Pinocchio 3x version, not yet online.

```cpp
//
// Copyright (c) 2021-2022 INRIA
//

#ifndef __pinocchio_python_math_multiprecision_boost_number_hpp__
#define __pinocchio_python_math_multiprecision_boost_number_hpp__

#include "pinocchio/math/multiprecision.hpp"

#include <boost/python.hpp>
#include <boost/python/return_value_policy.hpp>
#include <eigenpy/user-type.hpp>
#include <eigenpy/ufunc.hpp>
#include <sstream>

namespace
{

  template<class Backend>
  struct get_backend_precision {};

  template <unsigned Digits10, ::boost::multiprecision::mpfr_allocation_type AllocateType>
  struct get_backend_precision<::boost::multiprecision::mpfr_float_backend<Digits10,AllocateType>>
  {
    enum { value = Digits10 };
  };

}

namespace eigenpy
{
  namespace internal
  {
    template <class Backend, ::boost::multiprecision::expression_template_option ExpressionTemplates>
    struct getitem<::boost::multiprecision::number<Backend,ExpressionTemplates>>
    {
      
      typedef ::boost::multiprecision::number<Backend,ExpressionTemplates> Scalar;

      static PyObject* run(void* data, void* /* arr */) {
        Scalar & mpfr_scalar = *static_cast<Scalar*>(data);
        Backend & backend = mpfr_scalar.backend();
       
        if(backend.data()[0]._mpfr_d == 0) // If the mpfr_scalar is not initialized, we have to init it.
        {
          mpfr_scalar = Scalar(0);
//          unsigned int precision = get_backend_precision<Backend>::value ? get_backend_precision<Backend>::value : backend.default_precision();
//          mpfr_init2(backend.data(), ::boost::multiprecision::detail::digits10_2_2(precision));
        }
        bp::object m(boost::ref(mpfr_scalar));
        Py_INCREF(m.ptr());
        return m.ptr();
      }
    };
  }

[...]

PR #370, opened 2023-05-18
custom numeric type: Example and two bugfixes

ofloveandhate:
# Example: custom numeric type — and fixes for the two crash classes it exposes

This PR has grown from an example into example + fixes. It now contains:

1. **A standalone example** (`examples/custom_numeric_type/`) binding a Boost.Multiprecision custom numeric type — variable-precision `mpfr_float` and `mpc_complex` — through eigenpy, as discussed in #365. It builds independently against an installed eigenpy and is not part of eigenpy's build or CI (no mpfr/GMP/MPC dependency is added to eigenpy).
2. **A fix for crashes on never-constructed user-dtype slots** (the reason the example's `np.zeros((3,4), dtype=...)` lines were commented out for years).
3. **A fix for a converter-storage overrun** for writable `Eigen::Ref` parameters passed by const reference, which corrupts the adjacent argument.

## Bug A: zero-filled but never-constructed user-dtype slots

eigenpy registers user types with `NPY_NEEDS_INIT`: numpy zero-fills fresh buffers (`np.zeros`, `np.empty`, ufunc outputs) but never runs a C++ constructor on the elements. Every dtype slot function and ufunc loop then reads those raw zero bytes as a constructed `T`. For most scalars all-zero *is* a valid 0 and nothing happens; for pointer-backed scalars such as Boost.Multiprecision mpfr/mpc numbers, all-zero is the *uninitialized* state (null `_mpfr_d` limb pointer) and any use crashes inside libmpfr (`np.dot`/`np.inner`, arithmetic on fresh arrays, `.astype()`/`.copy()`, `np.full`, `np.arange`, truthiness).

**Fix:** a new customization point, `eigenpy::user_type_traits<T>` (`include/eigenpy/user-type-traits.hpp`), with `requires_initialization` (default `false`) and `is_uninitialized(const T&)`. The slot functions in `user-type.hpp` and the loops in `ufunc.hpp` route reads through a guard that substitutes an exact `T(0)` for never-constructed slots (heal-in-place in `getitem`, zero-before-assign in `setitem`).

Key properties:
- **Defaulted trait compiles to the previous code.** `requires_initialization` is a compile-time constant; for every existing user type the guards dead-code-eliminate and the original code path is kept verbatim (including `dotfunc`'s Eigen `v0.dot(v1)`, conjugating semantics preserved exactly in the guarded loop). Header-only, no ABI change.
- Opt-in types must tolerate `operator=` into zeroed storage (BMP types do; this was already implied by `NPY_NEEDS_INIT` registration).

**Tests:** `unittest/user_type_uninitialized.cpp` registers a dependency-free `GuardedScalar` whose all-zero state is an explicit invalid sentinel (mimicking mpfr) and counts every read of such a slot; `unittest/python/test_user_type_uninitialized.py` exercises creation/arithmetic/comparisons/matmul/dot/cast/copy/fill/nonzero paths and asserts zero violations. Without the guards the test fails immediately; the mpfr example (its pytest and `examplescript.py`) is the live multiprecision proof — its previously-crashing direct-creation lines are now enabled and passing.

## Bug B: writable `Eigen::Ref` by const reference overruns its converter storage

`eigen-from-python.hpp` specialized `boost::python::detail::referent_storage` only for `Eigen::Ref<MatType>&` and `const Eigen::Ref<const MatType>&`. A **writable Ref passed by const reference** (`const Eigen::Ref<MatType>&`, non-const `MatType`) fell through to Boost.Python's generic storage, sized for the `Ref` alone — but eigenpy's converter placement-news the full `referent_storage_eigen_ref` bookkeeping struct (Ref + 3 pointers) into it. Measured on aarch64: 24 bytes allocated, 64 written (40-byte overrun into the adjacent argument's converter on the invoke frame). The generic destructor also never ran `~StorageType`: no write-back to the numpy array and a leaked `pyArray` reference.

On x86_64 with AVX it is worse: `EIGEN_DEFAULT_ALIGN_BYTES` is 32, so the specialized layouts place the storage member at offset 32 while the generic one has it at 16 — `eigen_from_py_construct`'s cast then *displaces* the write as well. This matches the ABI-sensitivity of the corruption observed downstream (adjacent scalar arriving with precision 0 → `mpc_set_prec(0)` abort).

**Fix:** add the missing `referent_storage` and `rvalue_from_python_data` specializations for `Eigen::Ref<MatType, Options, Stride> const&` (partial ordering keeps the existing const-const form preferred where it applies), plus a `static_assert` tripwire in `eigen-allocator.hpp` documenting the storage-size invariant.

**Tests:** `probeWritableRefByValue*` / `probeWritableRefByConstRef*` in `unittest/eigen_ref.cpp` pass a writable Ref (both forms, real and complex, contiguous and non-contiguous arrays) alongside two scalar arguments and assert the scalars arrive intact and write-back works. Without the fix the const-ref probes fail on aarch64 already; x86_64 CI exercises the layout that crashed downstream.

## Notes

- Verified locally on linux-aarch64 (clean micromamba env: Eigen 3.4.0, Boost 1.8x, numpy 2.x, Python 3.12): full ctest suite green, example script + pytest green including the previously-crashing operations.
- The example's `cmake_minimum_required` was bumped to 3.22 (required by current jrl-cmakemodules).

Fixes the crash classes discussed in #365.

## CI

The full workflow matrix was run on the fork before updating this PR: [ofloveandhate/eigenpy#1](https://github.com/ofloveandhate/eigenpy/pull/1) — 33 checks green (ubuntu incl. `-mavx2` Release/Debug, macOS ARM/Intel, Windows, ROS, Nix; the only failures are an apt-mirror 404 for `python3-pil` on the runner image, unrelated to this diff).


jcarpent (commented):
Thanks @ofloveandhate for the PR. I will make a pass on it when I have time in order to align it with the rest of the project. Is it fine for you?

jcarpent:
TODO:

- [ ] Remove useless files
- [ ] SImplify cmake usage
- [ ] Extend Python test with the creation of Numpy arrays used by a C++ program

ofloveandhate:
I would love to be able to build upon this to work on some problems I'm seeing with custom numeric types in EigenPy, and make this example more complete, but I am finding the setup stuff is a barrier.  Thanks!

ofloveandhate:
I'm back for more help, and to help this example of using Eigenpy with a custom numeric type become fully functional.  

I just added some code that makes EigenPy / the example library crash.  I think my code is missing something simple.  The failing code in Python is merely this:

```
import eigenpy_example_custom_numeric_type as example
import numpy as np

M = np.zeros((3,4),dtype=example.MpfrComplex)  # make an array of the custom numeric type
```

The problem is clearly uninitialized data, as per the assertion fail from MPC:

```
Assertion failed: (m_data[0].re[0]._mpfr_d), function data, file mpc.hpp, line 346.
```

(This is a step towards another problem I have with alignment)

---

I would love some help so that I can make arrays with my custom type in Numpy, and then feed them to C++ routines.  I think this PR is a good place to do it, so that the resulting example code is contributed directly to EigenPy.  (I am sorry it took me so long to follow up -- I switched jobs, moved continents, and am now in a full time research position 🥳)

ofloveandhate:
I'm getting closer to the problem I'm trying to identify:

> how to be able to make numpy containers of custom numeric type, via EigenPy, without converting from existing.

This most recent set of commits shows that I can make numpy array of custom dtype (MpfrFloat and MpfrComplex) IF I FIRST make an array in built-in type, and convert to my own type.

```
A = np.array( np.empty( (3,4)).astype(np.int64),dtype=num_type) 
# this is used in the user_type example in EigenPy
```

But, the following still fails, which feels like a showstopper:

```
B = np.zeros( (4,5), dtype=num_type)
```

What magic sauce am I still missing to be able to directly make numpy arrays of custom type?

ofloveandhate:
This PR now encompasses the fixes to the two bugs, as well as a standalone example library with a user type that requires initialization.

PR #393, opened 2023-10-25
Topic/override std def

### LINKED ISSUE #392: Python bindings changing type depending on imports
Hello all, I've found what I think is an issue that I would like your comments, please.

## Bug description

In a python script, I am trying to import both pinocchio and ompl, which I am able to. The strange behavior that I am witnessing is that, depending on which one I import first, the types of the variables within pinocchio AND ompl change! 

## Expected behavior

I was expecting them to don't change.

## Reproduction steps

When I run 

```py
from os.path import abspath, dirname, join
from sys import argv

from ompl import base, geometric
import pinocchio

# This path refers to Pinocchio source code but you can define your own directory here.
pinocchio_model_dir = join(dirname(dirname(str(abspath(__file__)))), "models")

# You should change here to set up your own URDF file or just pass it as an argument of this example.
urdf_filename = (
    pinocchio_model_dir
    + "/example-robot-data/robots/ur_description/urdf/ur5_robot.urdf"
    if len(argv) < 2
    else argv[1]
)

# Load the urdf model
model = pinocchio.buildModelFromUrdf(urdf_filename)
# print("model name: " + model.name)

# Create data required by the algorithms
data = model.createData()

print("Pinocchio model.names type:", end=" ")
print(type(model.names))
print("")


bounds = base.RealVectorBounds(model.nq)
print("OMPL base.RealVectorBounds.high type:", end=" ")
print(type(bounds.high))
print("")
```

I expected the output to be 

```
Pinocchio model.names type: <class 'pinocchio.pinocchio_pywrap.StdVec_StdString'>

OMPL base.RealVectorBounds.high type: <class 'ompl.util._util.vectorDouble'>
```

Yet the actual output is 

```
Pinocchio model.names type: <class 'ompl.util._util.vectorString'>

OMPL base.RealVectorBounds.high type: <class 'ompl.util._util.vectorDouble'
```

and if switch from 

```py
from ompl import base, geometric
import pinocchio
```

to

```py
import pinocchio
from ompl import base, geometric
```

the output becomes

```
Pinocchio model.names type: <class 'pinocchio.pinocchio_pywrap.StdVec_StdString'>

OMPL base.RealVectorBounds.high type: <class 'pinocchio.pinocchio_pywrap.StdVec_Double'
```

### System

- OS: ubuntu 22.04
- Pinocchio version: 2.6.20

jorisv:
Fix #392

### DIFF
--- a/include/eigenpy/std-vector.hpp
+++ b/include/eigenpy/std-vector.hpp
@@ -352,6 +352,44 @@ struct contains_vector_derived_policies
     return contains_algo<key_type>::run(container, key);
   }
 };
+
+///
+/// \brief Add standard method to a std::vector.
+/// \tparam NoProxy When set to false, the elements will be copied when
+/// returned to Python.
+///
+template <typename Container, bool NoProxy, typename CoVisitor>
+struct AddStdMethodToStdVector
+    : public boost::python::def_visitor<
+          AddStdMethodToStdVector<Container, NoProxy, CoVisitor> > {
+  typedef StdContainerFromPythonList<Container, NoProxy>
+      FromPythonListConverter;
+
+  AddStdMethodToStdVector(const CoVisitor &co_visitor)
+      : m_co_visitor(co_visitor) {}
+
+  template <class Class>
+  void visit(Class &cl) const {
+    cl.def(m_co_visitor)
+        .def("tolist", &FromPythonListConverter::tolist, bp::arg("self"),
+             "Returns the std::vector as a Python list.")
+        .def("reserve", &Container::reserve,
+             (bp::arg("self"), bp::arg("new_cap")),
+             "Increase the capacity of the vector to a value that's greater "
+             "or equal to new_cap.")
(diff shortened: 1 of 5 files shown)


include/eigenpy/std-vector.hpp
```diff
@@ -352,6 +352,44 @@ struct contains_vector_derived_policies
...
   }
 };
+
+///
+/// \brief Add standard method to a std::vector.
+/// \tparam NoProxy When set to false, the elements will be copied when
+/// returned to Python.
+///
+template <typename Container, bool NoProxy, typename CoVisitor>
+struct AddStdMethodToStdVector
```
jcarpent, line 362:
```suggestion
struct ExposeStdMethodToStdVector
```
might be more exprisse.

jcarpent (changes requested):
Thanks @jorisv for this nice PR.
We might just change the naming of the visitor.

PR #399, opened 2023-11-13
core: fix StdVectorPythonVisitor::expose API

jorisv:
Avoid overload issue between

```cpp 
static void expose(const std::string &class_name, const std::string &doc_string);
```
 and

```cpp
  static void expose(const std::string &class_name, const Visitor &visitor);
```

### DIFF
--- a/include/eigenpy/registration_class.hpp
+++ b/include/eigenpy/registration_class.hpp
@@ -30,6 +30,12 @@ class registration_class {
     return *this;
   }
 
+  template <class DerivedVisitor>
+  self& def(bp::def_visitor<DerivedVisitor> const& visitor) {
+    static_cast<DerivedVisitor const&>(visitor).visit(*this);
+    return *this;
+  }
+
   /// \see boost::python::class_::def(char const* name, F f)
   template <class F>
   self& def(char const* name, F f) {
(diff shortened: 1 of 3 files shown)


hrp2-14:

include/eigenpy/registration_class.hpp
```diff
@@ -30,6 +30,12 @@ class registration_class {
     return *this;
   }
 
+  template <class DerivedVisitor>
+  self& def(bp::def_visitor<DerivedVisitor> const& visitor) {
+    static_cast<DerivedVisitor const&>(visitor).visit(*this);
```
ManifoldFR, line 35: Is the downcast actually required? I thought boost::python::def_visitor was a crtp base class and already did the downcast.

ManifoldFR: it does, see "boost/python/def_visitor.hpp" line 78

jorisv: Unfortunatly, the method `DerivedVisitor const& def_visitor<DerivedVisitor>::derived_visitor() const` is protected.

And ` void def_visitor<DerivedVisitor>::visit(classT& c) const` is private :(

ManifoldFR: Oh I think you told me, why would they even do that...

jcarpent (approved):

PR #401, opened 2023-11-14
ci: add ci step to check changelog update before merging in master

jorisv:
Check if a changelog if updated before allowing a merge request


jcarpent (approved):

nim65s (approved):
Good idea, but I guess you should start to show an example, by updating the CHANGELOG in this PR  then :)

PR #375, opened 2023-06-20
Enhance support of Tensor module in external applications

jcarpent:

PR #403, opened 2023-11-16
Tensor support

jcarpent:


.github/workflows/windows-conda.yml
```diff
@@ -47,9 +47,14 @@ jobs:
...
+        set OLD_PATH=%PATH%
+        set PATH=%PATH%;%CD%
         :: Build and Install
         cmake --build . --config Release --target install
         if errorlevel 1 exit 1
+        :: Set old path
+        set PATH=%OLD_PATH%
 
         :: Build stubs
         git clone https://github.com/jcarpent/pybind11-stubgen.git
```
jorisv, line 60: I think the manual stub generation is not mandatory anymore.

Maybe, a better fix should be to modify the path in jrlcmakemodule when calling the stub generator.

jcarpent:
@jorisv Thanks for finding this issue. Could you integrate an optional argument in the stubs function to automatically inform the PATH variable on Windows?

jcarpent:
COuld you make this change in #375?

Issue #405, opened 2023-11-22
eigenpy version set to 0.0.0 while forgetting --recursive option


olivier-stasse:
Symptoms
========
Pinocchio not compiling generating a mistake stating that eigenpy does not have the proper version.
The system detects a 0.0.0 version for eigen.

Steps to reproduce the bug
=====================

On a 22.04 LTS :
```
cd /tmp
mkdir -p /tmp/test_ws/src
cd /tmp/test_ws/src
git clone https://github.com/stack-of-tasks/eigenpy.git
cd ..
colcon build --merge-install --packages-select eigenpy
cd src
git clone https://github.com/stack-of-tasks/pinocchio.git --recursive
cd ..
colcon build --merge-install --packages-select pinocchio
```
Output:
```
Starting >>> pinocchio
--- stderr: pinocchio
CMake Error at cmake/package-config.cmake:110 (find_package):
  Could not find a configuration file for package "eigenpy" that is
  compatible with requested version "2.7.10".

  The following configuration files were considered but not accepted:

    /tmp/test_ws/install/lib/cmake/eigenpy/eigenpyConfig.cmake, version: 0.0.0

Call Stack (most recent call first):
  CMakeLists.txt:167 (ADD_PROJECT_DEPENDENCY)


---
Failed   <<< pinocchio [4.45s, exited with code 1]
```

Of course putting the recursive flag when cloning eigenpy is fixing the matter. So there is no urgency in this matter.

Unfortunately it means that the lines https://github.com/stack-of-tasks/eigenpy/blob/2a86f649c079c25bc5c72cda117e79316d532e6e/CMakeLists.txt#L24 - https://github.com/stack-of-tasks/eigenpy/blob/2a86f649c079c25bc5c72cda117e79316d532e6e/CMakeLists.txt#L30
are not working properly.

To be honest, I would prefer the error in https://github.com/stack-of-tasks/eigenpy/blob/2a86f649c079c25bc5c72cda117e79316d532e6e/CMakeLists.txt#L20 

I guess that you did it for a good reason, unfortunately i did not have time to investigate why the former lines are not working properly.

olivier-stasse:
and thanks for the amazing work !

This issue filling is mostly to document this unfortunate behavior and saving time for other users.

nim65s:
Closing in favor of https://github.com/jrl-umi3218/jrl-cmakemodules/issues/640

If this is not enough, I'll reopen.

[...]

PR #406, opened 2023-11-22
CMake: check version is not 0.0

nim65s:
Hi,

This add a regression test for #405. For now, it shouldn't pass.


nim65s:
and https://github.com/stack-of-tasks/eigenpy/actions/runs/6955692656/job/18924999953?pr=406 is correctly succeeding

PR #407, opened 2023-11-23
[unittest] change check for Cpp17 when compiling std::optional test

ManifoldFR:
This PR:

- switches to checking for the `cxx_std_17` [CMake compile feature](https://cmake.org/cmake/help/latest/manual/cmake-compile-features.7.html) before adding the test for std::optional
- adds said compile feature to the `bind_optional_std` unittest target

This change requires CMake>=3.8.

My reasoning is that previously we checked for `17 <= CMAKE_CXX_STANDARD < 98` which is i) not really semantic ii) error-prone if `CMAKE_CXX_STANDARD` is not set by the user, by the environment variables or by CMake itself.  
In contrast, in the source code, the preprocessor would resolve the `EIGEN_WITH_CXX17_SUPPORT` independently check and compile support for std::optional anyway.

### DIFF
--- a/unittest/CMakeLists.txt
+++ b/unittest/CMakeLists.txt
@@ -49,13 +49,16 @@ function(config_bind_optional tagname opttype)
   configure_file(python/test_optional.py.in
                  ${CMAKE_CURRENT_BINARY_DIR}/python/${py_file})
   add_lib_unit_test(${MODNAME})
+  message(STATUS "Adding test for ${opttype}")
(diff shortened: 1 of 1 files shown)


jcarpent (approved):
Could you update the changelog?

ManifoldFR:
Ok, it seems from the CI that Ubuntu 20's GCC 9.4 has default compiler level C++14 whereas CMake detects C++17 is available.

PR #412, opened 2023-11-30
Expose std::array types

### LINKED ISSUE #411: Add support for std::array
We currently only support `std::vector<T, Allocator>` vector types and `std::map<K, V>`.

ManifoldFR:
**Remarks**

Slices (i.e. using `object[a:b]` in Python on a subscriptable object) of `std::array<Eigen::Matrix, K>` types do not work. They do work for other types - this comes from exposing the Eigen std::array types with `eigenpy::details::overload_base_item_for...` which redefines `::base_get_item` for the underlying indexing_suite visitor that is used.

I think they might not work for slices of `std::vector<Eigen::Matrix>` types either, no @jorisv ?

**Meta**

Resolves #411

### DIFF
--- a/include/eigenpy/std-array.hpp
+++ b/include/eigenpy/std-array.hpp
@@ -81,7 +81,9 @@ class array_indexing_suite
   }
 };
 
-template <typename array_type, bool NoProxy = false>
+template <typename array_type, bool NoProxy = false,
+          class SliceAllocator =
+              std::allocator<typename array_type::value_type> >
 struct StdArrayPythonVisitor {
   typedef typename array_type::value_type value_type;
   /// Fixed size of the array, known at compile time
@@ -107,7 +109,7 @@ struct StdArrayPythonVisitor {
       cl.def(bp::init<const array_type &>(bp::args("self", "other"),
                                           "Copy constructor"));
 
-      array_indexing_suite<array_type, NoProxy> indexing_suite;
+      array_indexing_suite<array_type, NoProxy, SliceAllocator> indexing_suite;
       cl.def(indexing_suite);
     }
   }
@@ -120,8 +122,10 @@ void exposeStdArrayEigenSpecificType(const char *name) {
   oss << "StdArr";
   oss << Size << "_" << name;
   typedef std::array<MatrixType, Size> array_type;
-  StdArrayPythonVisitor<array_type>::expose(
-      oss.str(), details::overload_base_get_item_for_std_vector<array_type>());
+  StdArrayPythonVisitor<array_type, false,
+                        Eigen::aligned_allocator<MatrixType> >::
+      expose(oss.str(),
+             details::overload_base_get_item_for_std_vector<array_type>());
 }
 
 }  // namespace eigenpy
--- a/include/eigenpy/std-array.hpp
+++ b/include/eigenpy/std-array.hpp
@@ -102,7 +102,7 @@ struct StdArrayPythonVisitor {
   template <typename DerivedVisitor>
   static void expose(const std::string &class_name,
                      const std::string &doc_string,
-                     const bp::def_visitor<DerivedVisitor> &) {
+                     const bp::def_visitor<DerivedVisitor> &visitor) {
     if (!register_symbolic_link_to_registered_type<array_type>()) {
       bp::class_<array_type> cl(class_name.c_str(), doc_string.c_str());
       cl.def(bp::init<const array_type &>(bp::args("self", "other"),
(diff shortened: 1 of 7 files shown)


hrp2-14:

include/eigenpy/std-vector.hpp
```diff
@@ -242,7 +242,6 @@ namespace eigenpy {
 template <typename vector_type, bool NoProxy>
 struct StdContainerFromPythonList {
   typedef typename vector_type::value_type T;
-  typedef typename vector_type::allocator_type Allocator;
```
jcarpent, line 245: Why this deletion?

include/eigenpy/std-array.hpp
```diff
@@ -0,0 +1,140 @@
...
+    if (from > to) {
+      return;
+    } else {
+      std::copy(first, last, container.begin() + from);
+    }
+  }
+
+  static bp::object get_slice(Container &container, index_type from,
+                              index_type to) {
+    if (from > to) return bp::object(std::array<data_type, 0>());
```
jorisv, line 74: This function has two case :
- In one case, `get_slice` return a std::array
- In the nominal case, `get_slice` return a std::vector

I think this behavior could be misleading. In some case, it will be possible to modify the slice length, and in other case, it will not be possible.

include/eigenpy/std-array.hpp
```diff
@@ -0,0 +1,140 @@
...
+
+  template <class Iter>
+  static void set_slice(Container &container, index_type from, index_type to,
+                        Iter first, Iter last) {
+    if (from > to) {
+      return;
+    } else {
+      std::copy(first, last, container.begin() + from);
+    }
+  }
```
jorisv, line 70: `set_slice` behavior in python is not exactly the same.

Elements between `from` and `to` are deleted. This mean, the container can be resized, and that not possible for std::array.

On the first `set_slice`, I think your version make sens. But, it can also be misleading when working in Python.

In the second version, you should ensure that `first`, `last` fit the array size.
The following code in the unit test create an invalid write : `ints[:] = [1, 2, 3, 4]`

include/eigenpy/std-array.hpp
```diff
@@ -0,0 +1,140 @@
...
+  typedef std::vector<data_type, SliceAllocator> slice_vector_type;
+  static constexpr std::size_t Size = std::tuple_size<Container>{};
+
+  template <class Class>
+  static void extension_def(Class &) {}
+
+  template <class Iter>
+  static void extend(Container &, Iter, Iter) {}
+
+  static void append(Container &, data_type const &) {}
```
jorisv, line 46: Since you have redefined extension_def, you don't have to redefine these two functions.

include/eigenpy/std-array.hpp
```diff
@@ -0,0 +1,140 @@
...
+  // no-op
+  static void delete_item(Container &, index_type) {}
+  // no-op
+  // no-op
+  static void delete_slice(Container &, index_type, index_type) {}
```
jorisv, line 51: Once again, this can be misleading in Python.

The developer use the delete function, but nothing happen. I prefer to cast an exception.

jorisv (changes requested):
The C++ code is perfect, but I think that the Python behavior can be misleading.

Before addressing the issues, we can maybe check what is done in other binding tools like pybind11.

Also, another approach is to directly convert std::array into an std::vector. This will address all the issues but will prevent to modify the std::array content from Python...

include/eigenpy/std-array.hpp
ManifoldFR: You're right. Ideally we would return some type of container that has dynamic (runtime-known) size but isn't resizable.

include/eigenpy/std-array.hpp
ManifoldFR: Resolved in f4d61cc

include/eigenpy/std-array.hpp
ManifoldFR: Thinking about it further, I think the best option would be to return a boost::span object (since we are not C++20 yet and don't have std::span). Which means exposing boost::span.

jorisv (approved):

ManifoldFR:
Okay, I ended up adding an `arr.tolist()` method as discussed with @jorisv.

jorisv (approved):

ManifoldFR:
Huh, some macos pipelines in the CI failed due to the miniconda setup.

PR #422, opened 2024-01-09
Fix ccache

jorisv:
- Use the right key to update ccache cache at each run
- Add ccache to windows
- Modernize the ci steps
- Use conda compiler


jcarpent (dismissed):

jcarpent (approved):
Thanks for the nice update and overall clean up.

PR #430, opened 2024-01-30
Add `boost::variant` support

jorisv:
- Add `BoostVariantConverter` to register a `boost::variant` conversion to an alternatives and alternatives conversion to `boost::variant`.
- Add `ReturnInternalBoostVariant` to return an alternative reference

Also:
- Cancel CI workflow when a new commit is pushed

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,3 @@
+BasedOnStyle: Google
+SortIncludes: false
+Standard: Cpp03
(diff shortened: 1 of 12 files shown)


ManifoldFR:
Is it possible to handle other variant types ? Like we did with the PR for optional types where we have a mechanism for both `boost::optional` and `std::optional` and their respective null/none values.

jcarpent (approved):

jorisv:
@ManifoldFR I will check if it's easy to handle std and boost with the same code

jcarpent:
std::variant are slow ;)

ManifoldFR (commented):
Just some spelling changes to do

PR #441, opened 2024-02-17
linters: add ruff & isort

nim65s:


stephane-caron:
Ruff can carry out `isort` by selecting "I001":

```toml
[tool.ruff]
select = [
    # pyflakes
    "F",
    # pycodestyle
    "E",
    "W",
    # isort
    "I001",
    # pydocstyle
    "D"
]
```

So we might as well only add ruff?

nim65s:
in that case we can also ask ruff to replace black

jcarpent (approved):

PR #444, opened 2024-03-01
Add pixi support

jorisv:
- [ ] pixi.toml
- [ ] CI
- [ ] README.md

### DIFF
--- a/pixi.toml
+++ b/pixi.toml
@@ -1,0 +1,37 @@
+[project]
+name = "eigenpy"
+version = "3.4.0"
+description = "Bindings between Numpy and Eigen using Boost.Python"
+channels = ["conda-forge"]
+platforms = ["win-64", "linux-64", "osx-64", "osx-arm64"]
+license = "BSD-2-Clause"
+license-file = "LICENSE"
+
+[tasks]
+clean = { cmd = "rm -rf build" }
+prepare = """cmake -G 'Ninja'
+                   -B build
+                   -S .
+                   -DCMAKE_CXX_COMPILER_LAUNCHER=ccache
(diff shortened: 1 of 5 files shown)


pixi.toml
```diff
@@ -0,0 +1,37 @@
+[project]
+name = "eigenpy"
+version = "3.4.0"
```
nim65s, line 3: We'll probably want to bump this in the `release` cmake target too

jorisv: Sure, do you want to do it ?

I'm not sure to merge this PR, I will only do it if I have positive feedback about this tool.

nim65s: I probably won't personaly use this tool, but if we think this tool is going to be used and we want to set it up on our other repos, we'll have to be sure that this version number stays in sync.

jorisv: Fixed in https://github.com/jrl-umi3218/jrl-cmakemodules/pull/684

jorisv:
We have to manage the following issue #508

olivier-roussel (approved):

.github/workflows/macos-linux-windows-conda.yml
ManifoldFR, line 1: This means that now we have a single workflow file for conda environments?

ManifoldFR (approved):
Looks good to me

.github/workflows/macos-linux-windows-conda.yml
jorisv: We have a single workflow file that manage Mac/Linux/Windows with conda packages

jcarpent (approved):

olivier-roussel (approved):

Issue #447, opened 2024-03-15
Numpy 2.0.0b1 is not supported


duburcqa:
Numpy 2.0.0 has been released on beta a few days ago and I waited to check how bad the migration would be. It turns out it is not working out of the box but supporting both 1.x and 2.x should be quite manageable. For instance, a one-liner is enough to fix `Boost::Python` (https://github.com/boostorg/python/issues/431) on 1.84 release, but I think it should be the same on 1.76.

For eigenpy, it is more tricky. I tried to do it myself, and for now I'm stuck here:
```
/Users/alexis.duburcq/workspace/src/jiminy/eigenpy/src/numpy.cpp:17:10: error: use of undeclared identifier 'PyArray_TypeNumFromName'
  return PyArray_TypeNumFromName(const_cast<char*>(type->tp_name));
```

Unfortunately, `PyArray_TypeNumFromName` has been removed with no alternative apparently. See [commit](https://github.com/seberg/numpy/blob/dde15a6320df672a40b3cfd0659de8fae7496cbc/numpy/_core/src/multiarray/arrayobject.c). I tried to backport the old implementation but it relies on some hidden global registry. I'm not sure it is possible to get access to it.

My patch so far:
```diff
diff --git a/include/eigenpy/numpy-allocator.hpp b/include/eigenpy/numpy-allocator.hpp
index 6165394..58f6415 100644
--- a/include/eigenpy/numpy-allocator.hpp
+++ b/include/eigenpy/numpy-allocator.hpp
@@ -138,7 +138,7 @@ struct numpy_allocator_impl_matrix<Eigen::Ref<MatType, Options, Stride> > {
                         outer_stride = reverse_strides ? mat.innerStride()
                                                        : mat.outerStride();
 
-      const int elsize = call_PyArray_DescrFromType(Scalar_type_code)->elsize;
+      const int elsize = PyDataType_ELSIZE(call_PyArray_DescrFromType(Scalar_type_code));
       npy_intp strides[2] = {elsize * inner_stride, elsize * outer_stride};
 
       PyArrayObject *pyArray = (PyArrayObject *)call_PyArray_New(
@@ -204,7 +204,7 @@ struct numpy_allocator_impl_matrix<
                         outer_stride = reverse_strides ? mat.innerStride()
                                                        : mat.outerStride();
 
-      const int elsize = call_PyArray_DescrFromType(Scalar_type_code)->elsize;
[...]

PR #448, opened 2024-03-15
Add support of numpy 2.0.0b1

duburcqa:
Fixes https://github.com/stack-of-tasks/eigenpy/issues/447

It requires a follow-up PR in Boost::Python itself [here](https://github.com/boostorg/python/pull/432), but it does not prevent from merging this PR.

### DIFF
--- a/src/register.cpp
+++ b/src/register.cpp
@@ -63,9 +63,9 @@ int Register::registerNewType(
     throw std::invalid_argument("PyType_Ready fails to initialize input type.");
   }
 
-  PyArray_Descr* descr_ptr =
-      new PyArray_Descr(*call_PyArray_DescrFromType(NPY_OBJECT));
-  PyArray_Descr& descr = *descr_ptr;
+  PyArray_DescrProto* descr_ptr = new PyArray_DescrProto();
+  Py_SET_TYPE(descr_ptr, &PyArrayDescr_Type);
+  PyArray_DescrProto& descr = *descr_ptr;
   descr.typeobj = py_type_ptr;
   descr.kind = 'V';
   descr.byteorder = '=';
@@ -98,7 +98,7 @@ int Register::registerNewType(
   PyArray_Descr* new_descr = call_PyArray_DescrFromType(code);
 
   if (PyDict_SetItemString(py_type_ptr->tp_dict, "dtype",
-                           (PyObject*)descr_ptr) < 0) {
+                           (PyObject*)new_descr) < 0) {
     throw std::invalid_argument("PyDict_SetItemString fails.");
   }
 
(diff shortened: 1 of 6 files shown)


jcarpent (changes requested):
Could you avoid renaming variables and perform the fix?

src/register.cpp
```diff
@@ -52,25 +52,25 @@ int Register::registerNewType(
...
       NPY_NEEDS_PYAPI | NPY_USE_GETITEM | NPY_USE_SETITEM | NPY_NEEDS_INIT;
-  descr.type_num = 0;
-  descr.names = 0;
-  descr.fields = 0;
-  descr.alignment =
-      alignement;  // call_PyArray_DescrFromType(NPY_OBJECT)->alignment;
+  proto.type_num = 0;
+  proto.names = 0;
+  proto.fields = 0;
+  proto.alignment =
```
jcarpent, line 68: Changing to `proto` does not improve the readability.

jcarpent: Please revert it.

jcarpent:
@duburcqa Thanks for the contribution to support the latest NumPy change. Yet, it seems important to first extend the support of EigenPy, then change some variable names if they improve the readability. Please make these two stages into two different PRs.

src/register.cpp
duburcqa: ok. I was following numpy naming convention for proto-description of arrays.

src/register.cpp
```diff
@@ -82,22 +82,22 @@ int Register::registerNewType(
...
   instance().type_to_py_type_bindings.insert(
       std::make_pair(type_info_ptr, py_type_ptr));
-  instance().py_array_descr_bindings[py_type_ptr] = new_descr;
-  instance().py_array_code_bindings[py_type_ptr] = code;
+  instance().py_array_descr_bindings[py_type_ptr] = descr;
+  instance().py_array_code_bindings[py_type_ptr] = type_num;
 
   //      PyArray_RegisterCanCast(descr,NPY_OBJECT,NPY_NOSCALAR);
-  return code;
+  return type_num;
```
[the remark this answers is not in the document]
duburcqa: yes, it is just wrong to consider it is a return code.

duburcqa:
I will fix support of `Python < 3.9` due to [unsupported Py_SET_TYPE](https://peps.python.org/pep-0674/).

duburcqa:
The CI is passing !

jcarpent (approved):
Thanks a lot @duburcqa. Very nice push forward !

jakirkham:
Thank you both for your hard work here! 🙏

Am curious, when building wheels for 3.5.0 or 3.5.1, were they built with NumPy 1 or 2?

duburcqa:
Numpy 1, there is still work to do to fully support numpy 2 in pre-compiled binaries.

jakirkham:
Ok thanks Alexis! 🙏

PR #496, opened 2024-08-27
Fix compatibility issue on Windows for NumPy 2.x

jcarpent:


jorisv:
This patch turn the `EIGENPY_ARRAY_API` symbol public again (like when building with numpy 1.2x).

I don't understand how this symbol is used by numpy API and I think this can create issue if a library using eigenpy want to use his own `XXX_ARRAY_API`. Anyway, we had the same issue with numpy 1.2x.

One way to avoid exposing this symbol would be to never call `PyArray_` function in our header file. But since we heavily use template, this should only be possible by wrapping all `PyArray_` function.

Some pinocchio unit test fail when linking against eigenpy + numpy 2.

jorisv:
Issue in Pinocchio seem related to a new way to manage conversion to user type to numpy array.

- In [numpy 2](https://numpy.org/doc/stable/user/basics.interoperability.html#the-array-method) the signature should be `__array__(self, dtype=None, copy=None)`
- In [numpy 1.2x](https://numpy.org/doc/1.26/user/basics.interoperability.html#the-array-method) the signature should be `__array__(self)`

jorisv:
@jcarpent I will make a test on Windows and if it's ok we can merge this PR.

jorisv:
@jcarpent This PR is also working on Windows.

We can merge it.

I will create an issue to add a CI job to build against numpy 1.2x but I will probably do that in the pixi PR.

PR #502, opened 2024-09-03
Add helper for type_info

jcarpent:


include/eigenpy/type_info.hpp
```diff
@@ -0,0 +1,77 @@
...
+    cl.def("std_type_info", &std_type_info, bp::arg("self"),
+           "Queries information of the type of *this as a std::type_index.");
+  }
+
+ private:
+  static boost::typeindex::type_index boost_type_info(const C& self) {
+    return boost::typeindex::type_id<C>();
+  }
+
+  static std::type_index std_type_info(const C& self) { return typeid(C); }
```
ManifoldFR, line 72: Shouldn't it be `return typeid(self)` ? In case the object is actually a subclass of `C` at runtime

src/type_info.cpp
```diff
@@ -0,0 +1,68 @@
...
+#include <typeindex>
+
+#include <boost/python.hpp>
+#include <boost/type_index.hpp>
+
+namespace bp = boost::python;
+
+void exposeStdTypeIndex() {
+  typedef std::type_index Self;
+  bp::class_<Self>(
```
ManifoldFR, line 15: You can also add a `pretty_name()` method here, using the lambda
```cpp
[] (const Self &value) -> std::string { return boost::core::demangle(value.name()); }
```

include/eigenpy/type_info.hpp
jcarpent: Thanks @ManifoldFR for the nice feedback. I've applied it.

src/type_info.cpp
jcarpent: Perfect. Thanks a lot for the nice suggestion.

include/eigenpy/type_info.hpp
```diff
@@ -0,0 +1,77 @@
...
+      "Returns information of the type of value as a "
+      "boost::typeindex::type_index (can work without RTTI).");
+  boost::python::def(
+      "boost_type_info",
+      +[](const T& value) -> boost::typeindex::type_index {
+        return boost::typeindex::type_id_runtime(value);
+      },
+      bp::arg("value"),
+      "Returns information of the type of value as a "
+      "boost::typeindex::type_index (can work without RTTI).");
```
ManifoldFR, line 38: Is exposing the function under those two names intended?

jcarpent: Proposing a default version

ManifoldFR (dismissed):
Looks good to me

PR #504, opened 2024-09-17
[std-map] add more general visitor for map types, fix non-default-constructible types

ManifoldFR:
+ test boost::unordered_map
+ pull StdMapPythonVisitor out of eigenpy::python namespace (shouldn't exist, add using-decl for backwards compatibility)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,13 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ## [Unreleased]
 
 ### Added
+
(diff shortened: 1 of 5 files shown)


ManifoldFR:
@edantec This generalises your previous PR in release 3.9.0, and fixes a namespace issue

ManifoldFR:
Interesting point in the latest commit: Boost.Python's map_indexing_suite does something annoying which is to assume the `value_type` in a Container type (a `std::pair<key_type, data_type>` for map types) is default-constructible. It also does that for vector_indexing_suite, and in both cases the `set_item()` does something like this:
```cpp
container[i] = v;
```
which is highly illegal.

jcarpent (approved):

PR #510, opened 2024-09-30
Restore cxx11 compatibility

jorisv:


ManifoldFR (dismissed):
man cxx11 was so bad

ManifoldFR (approved):
lgtm

PR #514, opened 2024-11-02
Fix linking to the wrong Python library for Debug built type on Windows

duburcqa:


jorisv:
Hello @duburcqa,

Thank for the PR.
When I read `wrap_python.hpp` and your PR title I'm a little confuse. Did this PR allow to:

- Build our Python modules in debug mode and be loaded by a Python release runtime on Windows ?
- Build our Python module in debug mode and be loaded by a Python debug runtime on Windows ?
- Both ?

duburcqa:
The point is rather linking with the Python library that Boost Python wants us to link with. The current implementation in eigenpy is by-passing Boost Python loading mechanism entirely because of this. Even worst, only compilation units that includes "include/eigenpy/numpy.hpp" are affected, but not the others, which means that the whole project may end up linking against multiple Python libraries at the same time.

But to answer for question, it allows both, depending on whether the pragma definition "BOOST_PYTHON_DEBUG" is set.

duburcqa:
> I don't really get why an header will change our linking behavior.

That is just how it is. This information can be found at several places in the documentation, for instance [here](https://live.boost.org/doc/libs/1_84_0/libs/python/doc/html/building/python_debugging_builds.html) or [here](https://www.boost.org/doc/libs/1_31_0/libs/python/doc/building.html).

jorisv:
Thank for the link. The first one explain well the issue.
I also find [this link](https://docs.python.org/3/using/configure.html#debug-build) that explain well what changed in the debug build and that since Python 3.8 there is no ABI incompatibility between debug and release Python build.
And the related [source code](https://github.com/python/cpython/blob/3966d8d626fc9adcdf0663024d4ae6e6be7ef858/PC/pyconfig.h.in#L311-L336).

So on **Windows**, pyconfig.h (included by Python.h) can call some MSVC specific pragma to link against a DLL.
It use the `_DEBUG` define to choose which library to link to. But this define is used commonly and we can then link against the wrong library.
To avoid this issue `boost/python/detail/wrap_python.hpp`  undef _DEBUG and define it back after including Python.h.

I was able to reproduce the issue and the patch fixed it.

I will modify the comment to precise it's a Windows specific issue.

jorisv (approved):

PR #538, opened 2025-03-26
Fix issue for sparse matrix conversions

jcarpent:
Very nice fix @ManifoldFR ;)

### DIFF
--- a/include/eigenpy/sparse/eigen-from-python.hpp
+++ b/include/eigenpy/sparse/eigen-from-python.hpp
@@ -1,5 +1,5 @@
 //
-// Copyright (c) 2024 INRIA
+// Copyright (c) 2024-2025 INRIA
 //
 
 #ifndef __eigenpy_sparse_eigen_from_python_hpp__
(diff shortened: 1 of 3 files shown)


include/eigenpy/sparse/eigen-from-python.hpp
```diff
@@ -156,6 +156,10 @@ void eigen_sparse_matrix_from_py_construct(
     }
     MapMatOrRefType sparse_map(m, n, nnz, indptr.data(), indices_ptr, data_ptr);
 
+#if EIGEN_VERSION_AT_LEAST(3, 4, 90)
+    sparse_map.sortInnerIndices();
```
ManifoldFR, line 160: Probably at some point we would like to add a check (and throw an exception) if the indices are unsorted and the Eigen version is below 3.4.90 ?

ManifoldFR:
I will open a PR on nanobind to address this issue ahead of our transition to it.

include/eigenpy/sparse/eigen-from-python.hpp
jcarpent: Yes, sure.

PR #545, opened 2025-04-25
Add user-defined literal `""_a` for `bp::arg`

ManifoldFR:
This PR adds a C++11 user-defined literal which is used as a shortcut for `bp::arg(name)`.

The same literal already exists in both pybind11 and nanobind. I also have a version of this for internal use in aligator's bindings.

Users can replace `bp::arg("truc")` by `"truc"_a`, after importing the operator into the current scope, or using the simpler `using namespace eigenpy::literals`.

## Usage:

```cpp
void my_fun(int x, int y, int z = 42)
{
    // do stuff
}
// ...
// in module definition:

using namespace eigenpy::literals;

bp::def("my_fun", my_fun, ("x"_a, "y", "z"_a = 42));
```


nim65s (dismissed):

jorisv (dismissed):

jcarpent (dismissed):

ManifoldFR:
In concert with @jorisv, we are using this PR to remove ubuntu 20.04 from the CI, and adding 24.04 instead.

I think you would approve @nim65s ? this would gel with your latest messages on matrix 😃

jorisv (approved):

nim65s (approved):

PR #546, opened 2025-04-25
Update C++ standard for clang-format to 11, apply updated formatting rules

ManifoldFR:
- **pre-commit config : actually delegate to .clang-format config file**
- **.clang-format : increase Standard to Cpp11**
- **pre-commit: run all**
- **.git-blame-ignore-revs : update**
- **Update CHANGELOG**

This PR would require setting the minimum required version of the C++ standard to 11 I think, because stuff like `vector<A<T>>` without a space between the last two angle brackets for templates doesn't work on some C++03 compilers if I recall correctly.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,3 +1,3 @@
 BasedOnStyle: Google
 SortIncludes: false
-Standard: Cpp03
+Standard: Cpp11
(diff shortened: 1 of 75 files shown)


nim65s (approved):

ManifoldFR:
I think we should merge this before #545. Then I can pop off the last commit off of that PR since we won't have to fight the formatting.

PR #591, opened 2025-09-10
Don't trigger CI on push and pull_request event

jorisv:
This will avoid building 2 time the same commit.
Also, try to homogenize workflows.

### DIFF
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -4,23 +4,7 @@ on:
   push:
     branches:
       - devel
-    paths-ignore:
-      - 'doc/**'
-      - '.gitlab-ci.yml'
-      - '.gitignore'
-      - '*.md'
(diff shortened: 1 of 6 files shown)


.github/workflows/nix.yml
```diff
@@ -3,17 +3,33 @@ name: "CI - Nix"
 on:
   push:
     branches:
-      - master
       - devel
+    paths-ignore:
```
nim65s: We had an issue with that here: https://github.com/coal-library/coal/pull/705
Not sure if it sitll apply

jorisv: Because of the `fileset` attribute in the flake.nix ?

nim65s: Something else was broken, because iirc the CI was just not running at all

ManifoldFR (approved):

PR #592, opened 2025-09-23
Add support for Python slice objects for `std::vector`

cmastalli:
The current  `__getitem__` only accepts integers, which means Eigenpy does not yet support Python slicing. So a slice object (xs[1:]) hits the "Invalid index type" path, as triggered in today's Crocoddyl pre-commit PR (see https://github.com/loco-3d/crocoddyl/actions/runs/17928428896/job/50980225223?pr=1433).

This PR introduces a slice-aware `__getitem__`. The basic strategy is to add an overload that accepts `boost::python::slice` and returns a Python list (with element references if you need mutability, mirroring our single-index path).

I also included a copyright update, but I'm happy to remove it if you guys feel it's not justified.


ManifoldFR:
Hi Carlos,
Thanks a lot for this! I'd added slice support for a type in one of my packages some time ago and I never realised we never had this in eigenpy.

I'll take care of reviewing this PR @jcarpent @nim65s.

@cmastalli can you add a test for this feature?

include/eigenpy/std-vector.hpp
```diff
@@ -104,6 +106,29 @@ struct overload_base_get_item_for_std_vector
     return bp::object(bp::handle<>(convert(*i)));
   }
 
+  static boost::python::object base_get_item_slice(
```
ManifoldFR, line 111: I think it's also possible to support list indexing (using `bp::list`), i.e.
```python
L = [0, 4, 5]
x = np.arange(1, 10)  # or a list
x[L]  # should be array([1, 5, 6])
```
Would you have time to do that?

include/eigenpy/std-vector.hpp
```diff
@@ -104,6 +106,29 @@ struct overload_base_get_item_for_std_vector
...
+      boost::python::back_reference<Container&> container,
+      boost::python::slice slice) {
+
+    namespace bp = boost::python;
+    Py_ssize_t start, stop, step, slicelen;
+    const Py_ssize_t n = static_cast<Py_ssize_t>(container.get().size());
+
+    if (PySlice_GetIndicesEx(slice.ptr(), n, &start, &stop, &step, &slicelen) != 0)
+      bp::throw_error_already_set();
```
ManifoldFR: I think you should use the `slice.get_indices(const RandomAccessIterator &begin, const RandomAccessInterator &end)` API instead, see
https://www.boost.org/doc/libs/latest/libs/python/doc/html/reference/object_wrappers/boost_python_slice_hpp.html. It's supposed to handle bound problems.

include/eigenpy/std-vector.hpp
cmastalli: In the new version, we are not only handling list indexing but also tuple indexing. See the unit tests!

include/eigenpy/std-vector.hpp
cmastalli: Done!

cmastalli:
@ManifoldFR -- please note that apart from handling your comments, I included a Cmake command to enable us to run the Python tests individually.

unittest/CMakeLists.txt
```diff
@@ -73,6 +73,15 @@ function(add_python_eigenpy_lib_unit_test name source)
...
   set_tests_properties(${test_target} PROPERTIES DEPENDS ${PYWRAP})
+  string(REPLACE "unittest/" "" source_rel "${source}")
+  add_custom_target(
+    "test-${name}"
+    ${CMAKE_COMMAND}
+    -E
+    env
+    PYTHONPATH=${PROJECT_BINARY_DIR}/lib:${PROJECT_BINARY_DIR}/python:$ENV{PYTHONPATH}
+    ${PYTHON_EXECUTABLE}
+    "${CMAKE_CURRENT_SOURCE_DIR}/${source_rel}")
```
jorisv:
```suggestion
```

You can remove this code. `add_python_unit_test` is already setting the appropriate PYTHONPATH to the test.

include/eigenpy/std-vector.hpp
```diff
@@ -104,6 +108,79 @@ struct overload_base_get_item_for_std_vector
     return bp::object(bp::handle<>(convert(*i)));
   }
 
+  static boost::python::object base_get_item_slice(
+      boost::python::back_reference<Container&> container,
+      boost::python::slice slice) {
+
+    namespace bp = boost::python;
```
ManifoldFR: I don't think this line is necessary, as there are uses of the `bp::` shorthand before that line and that code worked.

include/eigenpy/std-vector.hpp
ManifoldFR: Good, this aligns it with numpy arrays who can be indexed in both ways

(vanilla Python lists don't support list/tuple indexing but my suggestion was to allow it anyway).

include/eigenpy/std-vector.hpp
```diff
@@ -104,6 +108,79 @@ struct overload_base_get_item_for_std_vector
...
+      bp::throw_error_already_set();
+    }
+    return static_cast<index_type>(idx);
+  }
+
+  static bp::object elem_ref(Container& c, index_type i) {
+    typename bp::to_python_indirect<value_type&,
+                                    bp::detail::make_reference_holder> conv;
+    return bp::object(bp::handle<>(conv(c[i])));
+  }
```
ManifoldFR: Will this work with most types we bind? @jorisv

CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+- Supported Python slicing in std::vector in ([#592](https://github.com/stack-of-tasks/eigenpy/pull/592))
```
ManifoldFR:
```suggestion

### Added

- Support for Python slice, tuple and list indexing for `std::vector` bindings ([#592](https://github.com/stack-of-tasks/eigenpy/pull/592))
```

ManifoldFR (changes requested):

unittest/CMakeLists.txt
cmastalli: Thanks for the information! I just reverted it!

CHANGELOG.md
cmastalli: Done!

include/eigenpy/std-vector.hpp
cmastalli: Nice catch! I removed it.

cmastalli:
Hey @ManifoldFR  and @jorisv! I have addressed your comments; however, I am unable to make changes to my commits as I am using a MacBook (i.e., the default filesystem is case-insensitive).

In a nutshell, Git can’t represent the `accelerate.hpp` vs `Accelerate.hpp` files at once and gets "confused." This means that I cannot perform `git rebase -i` as it thinks the files are unstaged. I suggest (1) merging these two files to be compliant with macOS developers in another PR, and (2) squashing my commits when merging.

jorisv:
Hello @cmastalli,

You're right about accelerate. We keep the old `accelerate.hpp` header for back compatibility but didn't notice this is an issue for non case sensitive OS.

@ManifoldFR I see three options:
- We move `Accelerate.hpp` to `accelerate.hpp`. This preserve EigenPy API <= 3.11 API.
- We rename `Accelerate.hpp` into `AccelerateSomething.hpp`
- We remove `accelerate.hpp`. This break the API but since I don't think many people are using it, it's maybe not an issue…

ManifoldFR:
I say we go with the third option @jorisv. Accelerate isn't even something someone running a packaged version of an eigenpy release would be using - because the package would depend on Eigen 3.4 at most, which doesn't have Accelerate support! So breaking the include is a-ok in my book.

ManifoldFR:
Carlos can you fix the [NPY002](https://docs.astral.sh/ruff/rules/numpy-legacy-random/) error the ruff linter reports?

jorisv:
@ManifoldFR @cmastalli Accelerate.h has been fixed and I have rebase this PR on devel

cmastalli:
> Carlos can you fix the [NPY002](https://docs.astral.sh/ruff/rules/numpy-legacy-random/) error the ruff linter reports?

Done!

cmastalli:
@jorisv and @ManifoldFR -- I have squashed a few commits and fixed the NPY002 error. It could be good to merge now!

ManifoldFR:
Ruff and clang-format are still complaining in the pre-commit run. Did you have pre-commit installed in your checked-out repo?

ManifoldFR:
Otherwise it will be good to merge

cmastalli:
> Otherwise it will be good to merge

Awesome! I didn't know that Eigenpy is using pre-commit. I just pushed a commit that runs `pre-commit` linting.

ManifoldFR (approved):

PR #593, opened 2025-09-24
Remove `accelerate.hpp`

jorisv:
This header clash with `Accelerate.hpp` in non case sensitive OS.

We don't consider it as a API breaking change since Accelerate.hpp need Eigen devel and then is not a easy to use API.


ManifoldFR (approved):

ManifoldFR:
@cmastalli can you rebase your PR on the devel branch after this is merged? Thanks!

nim65s (approved):

